PR #11, opened 2015-04-07
Correct `@interval` macros.

dpsanders:


lbenet:
After merging #12, conflicts arose here. Can you solve them? I think they appear in REQUIRE and .travis.yml; maybe also in some of the tests files.

dpsanders:
I have rebased. Please let me add documentation and tests to fix things we have discussed.

dpsanders:
Do you think I should squash some commits?

lbenet:
Yeah; makes life easier :-)

lbenet:
Incidentally, running the test in julia 0.4 send a deprecation warning:
```julia
WARNING: int(x::FloatingPoint) is deprecated, use round(Int,x) instead.
```

I think this is a good time to require Compat.jl to get rid of this warning. What do you think?

Then I could merge and we look for whatever consequences.

dpsanders:
Sure, can you make a PR against my PR with the changes needed to include Compat.jl? I haven't used it very much.

dpsanders:
Or just use the following for the moment?
```
a = 3.1
convert(Int, floor(a))
```

lbenet:
I'll merge and add a parallel pull-request... Much easier :-)

dpsanders:
I would like to do the `rationalize` but leave a `BigFloat` as is. Does that sound like a good compromise?

lbenet:
Do the tests pass? If yes, I think it makes sense.

PR #12, opened 2015-04-08
Change from using Base.Test to using FactCheck

dpsanders:


lbenet:
Can you delete this branch, to avoid too much cluttering?

PR #16, opened 2015-04-10
Made tests into a package. Added FactCheck.exitstatus()

dpsanders:


lbenet:
Are tests passing?

dpsanders:
Yes. I have enabled Travis:

https://travis-ci.org/dpsanders/ValidatedNumerics.jl

dpsanders:
The point was to add `FactCheck.exitstatus()` because otherwise `Pkg.test()` reports that the tests are passing, even though they aren't!

PR #17, opened 2015-04-10
Use rationalize to convert Float64's into rationals,

dpsanders:
before converting them to intervals.

### DIFF
--- a/src/intervals/interval_macros.jl
+++ b/src/intervals/interval_macros.jl
@@ -12,7 +12,6 @@ end
 
 
 macro rounding(T, expr, rounding_mode)
-
     quote
         with_rounding($T, $rounding_mode) do
             $expr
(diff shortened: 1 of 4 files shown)


dpsanders:
Going back to the question about what to do if a `BigFloat` is provided, the `Interval` constructor already allows constructing a strictly thin interval:

    Interval(big(0.1))

So I suggest to widen the interval slightly by having

    x = big("0.1")
    @interval(x)

make `Interval(prevfloat(x), nextfloat(x))`

This would also be a faster alternative for `Float64`s (instead of having to use `rationalize`).

lbenet:
I agree with the last part, that this may be a faster alternative for `Floats64`s.

The question is, what would you like to have for `x=big(0.5)` and `@interval(x)`? My naive answer is a zero diameter interval. Yet, with the above scheme you would get something odd, since:
```julia
julia> prevfloat(big(0.5))
4.9999999999999994e-01 with 53 bits of precision

julia> nextfloat(big(0.5))
5.0000000000000011e-01 with 53 bits of precision
```

dpsanders:
Hmmm, I take your point. 

I think my idea was that if you're trying to construct an interval from a BigFloat, it's after a long calculation. If you are starting from 0.5, and you know it's exactly 0.5, then you can just do `@interval(0.5)` or even `@interval(1/2)` or `@interval(1)/2`.

On the other hand, if you're creating an interval from a `BigFloat` after a long calculation, it's likely that you've already lost precision and you should have done the whole calculation with intervals.

So maybe the correct solution is to prohibit constructing intervals from `Float64`s or `BigFloat`s altogether...

lbenet:
What I recall was that if an interval needed to be constructed using a `BigFloat`, then this was not going to be rounded, since (1) the user knew what he/she was doing, or (2) it was used from an already computed interval, so rounding was already taken care of.

> So maybe the correct solution is to prohibit constructing intervals from Float64s or BigFloats altogether...

Here, again, we may have the situation in which `f(x) = x + 0.1`. If 0.1 does not become an interval, we may be shifting the results without an actual control...

dpsanders:
Very good point. But then what about

    f(x) = x + BigFloat("0.1")

So we should be rounding a `BigFloat` as well (which, apparently, currently happens in a magic way that I  don't have time to check right noe).

lbenet:
Good point as well :-)

dpsanders:
Didn't mean to close this...

dpsanders:
Closed in favour of #18 .

PR #9, opened 2014-11-12
Changes in newton.jl

lbenet:
I have cleaned-up the code, changed the way of calling `N` (now
without any reference to f_prime), and changed the symbols associated
with the roots. I included an exit criterion related with the number
of levels that can be tried (for now, being 30).

PR #19, opened 2015-04-27
Changes in newton with tests, and improvements in interval_trig

lbenet:
This pull-request essentially does the same as #9 but with the updated version in master. In addition, it includes (previously commented) tests for the auto_diff module and new tests for newton. Some tests for newton uncovered an inconsistency in the type of the interval induced in interval_trig; I hope this doesn't create new problems, though tests pass.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,3 +2,4 @@ julia 0.3
 Docile
 FactCheck
 Compat
+Polynomials
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,4 @@
 julia 0.3
 Docile
-FactCheck
 Compat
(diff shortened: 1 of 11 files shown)


lbenet:
Tests for v0.3 are failing (v0.4 pass), since `parse(BigFloat,x)` with `x` an `AbstractString` is not being recognized. This should be solved once JuliaLang/Compat.jl/pull/76 is merged.

dpsanders:
I suggest just adding that to ValidatedNumerics for the moment.

lbenet:
Done... finally :-)

dpsanders:
Thanks! :)

PR #20, opened 2015-05-04
RFC: Make trigonometric functions rigorous by calculating correctly the quadrants

dpsanders:


dpsanders:
I think this is ready, @lbenet. Feel free to add more tests! Would you like me to rebase?

src/ValidatedNumerics.jl
```diff
@@ -6,7 +6,7 @@ using Compat
 
 @compat round
 @compat parsefloat
-
+@compat ifloor
```
lbenet: `ifloor` is deprecated in favor of `floor(Integer,x)`

src/intervals/interval_trig.jl
```diff
@@ -1,113 +1,141 @@
 
 #----- From here on, NEEDS TESTING ------
 
+
+const float_pi_interval = @floatinterval(pi)
+const big_pi_interval = @interval(pi)
```
lbenet: I guess there is some odd dependence `big_pi_interval` with regards to the default precision set (53). Probably all these constants should go into a type structure, which redefines them if the precision is changed.

src/root_finding/automatic_differentiation.jl
```diff
@@ -40,6 +40,7 @@ end
 
 sin(x::Jet) = Jet(sin(x.val), x.der*cos(x.val))
 cos(x::Jet) = Jet(cos(x.val), -x.der*sin(x.val))
+tan(x::Jet) = Jet(tan(x.val), x.der / cos(x.val)^2 )
```
lbenet: There is a `sec(x)` function, so this could be rewritten in a julian way as `Jet(tan(x.val), x.der * sec(x) ^2)`

dpsanders: Since `sec(x)` is implemented directly as `one(T)/cos(x)`, I don't think that anything is gained by this (and in fact it could even be slightly less efficient).

src/intervals/interval_trig.jl
dpsanders: Yes, I agree. I propose to do this in a separate PR.

PR #21, opened 2015-05-04
WIP: Implements two types of rounding: narrow/slow and wide/fast. The former changes the rounding mode as before; the latter does everything with RoundNearest and just takes prevfloat and nextfloat. This seems to be much (~30x) faster.

dpsanders:
`:narrow` changes the hardware rounding mode (for `Float64`) as before; `:wide` does everything with `RoundNearest` and just takes `prevfloat` and `nextfloat`. This seems to be much (~30x) faster.

The tests must be updated to use `roughly` from `FactCheck`.


dpsanders:
Closing in favor of #25

PR #22, opened 2015-05-05
Replaces big_transf and float_transf with a single make_interval function in interval_rounding.jl

dpsanders:


src/intervals/old_rationalize.jl
```diff
@@ -0,0 +1,29 @@
+# Version of `rationalize` from 0.3 that works correctly with rounding modes
```
lbenet: I think this comment should include a bit more details about the source file. Though the license is quite open, we should mention it somewhere. Or do you think this is worthless?

src/intervals/interval_rounding.jl
```diff
@@ -0,0 +1,193 @@
...
+- :narrow  -- changes the floating-point rounding mode to `RoundUp` and `RoundDown`.
+This gives the narrowest possible interval.
+
+- :wide -- Leaves the floating-point rounding mode in `RoundNearest` and uses
+`prevfloat` and `nextfloat` to achieve directed rounding. This creates an interval of width 2`eps`.
+""" ->
+
+get_interval_rounding() = INTERVAL_ROUNDING[end]
+
+set_interval_rounding(mode) = INTERVAL_ROUNDING[end] = mode  # a symbol
```
lbenet: I think it is a good idea to add checks to `mode` so it only accepts (for the time being) `:narrow` or `:wide`?

src/intervals/interval_rounding.jl
```diff
@@ -0,0 +1,193 @@
...
+
+        end
+    end
+end
+
+
+@doc doc"""`make_interval` is used by `@interval` to create intervals from individual elements of different types"""->
+make_interval(::Type{BigFloat}, x::String)    =  @thin_round(BigFloat, @compat parse(BigFloat,x))
+
+make_interval(::Type{BigFloat}, x::String)    =  @thin_round(BigFloat, BigFloat(x))
```
lbenet: I think this line was supposed to be replaced by the definition `make_interval(::Type{BigFloat}, x::String)` two lines above; it triggers a deprecation warning.

src/intervals/interval_rounding.jl
```diff
@@ -0,0 +1,193 @@
...
+make_interval(::Type{BigFloat}, x::Float64)   =  make_interval(BigFloat, rationalize(x))  # NB: converts a float to a rational -- this could be slow
+
+make_interval(::Type{BigFloat}, x::BigInt)    =  @thin_round(BigFloat, convert(BigFloat, x))
+
+make_interval(::Type{BigFloat}, x::BigFloat)  =  @thin_round(BigFloat, 1.*x)  # convert to possibly different BigFloat precision
+make_interval(::Type{BigFloat}, x::Interval)  =  @round(BigFloat, big(1.*x.lo), big(1.*x.hi))
+
+
+
+make_interval(::Type{Float64}, x::String)    =  @thin_round(Float64, parsefloat(x))
```
lbenet: `@floatinterval("0.1")` triggers a deprecation warning to use `parse(Float64,s)`.

test/interval_tests/loop_tests.jl
```diff
@@ -57,17 +57,50 @@ function calc_pi3(N)
...
+    sqrt(6*S5)
+end
+
 
 facts("Pi tests") do
 
+    big_pi = with_bigfloat_precision(256) do
+        big(pi)
+    end
+
```
lbenet: Shouldn't we make the tests related to pi, precision independent? 

I think, it is better to carry on the tests defining `big_pi` (in the lines above) as
```julia
big_pi = with_bigfloat_precision(get_bigfloat_precision()) do
    big(pi)
end
```

src/intervals/old_rationalize.jl
dpsanders: Agreed; I'll add it.

src/intervals/interval_rounding.jl
dpsanders: Agreed; I'll do so.

src/intervals/interval_rounding.jl
dpsanders: Yes, good catch, thanks!

src/intervals/interval_rounding.jl
dpsanders: Will change, thanks.

test/interval_tests/loop_tests.jl
dpsanders: Actually in this case I don't think that's necessary, since all it's used for is to test that it lies inside the interval generated by the sums.

dpsanders:
Finally tests passing and ready for merge @lbenet

lbenet:
Feel free to delete make_interval branch...

dpsanders:
Thanks :)

PR #25, opened 2015-05-13
RFC: Add IntervalParameters object; @interval creates intervals of types determined by those parameters

dpsanders:


lbenet:
Tests failed...

lbenet:
Tests failed...

dpsanders:
Tests now pass!

dpsanders:
Something strange is happening -- certain results of calculations with powers are different on Travis from on my machines. @lbenet could you check if the tests pass for you, please?

lbenet:
I'll take a look on this now.

src/intervals/interval_functions.jl
```diff
@@ -1,95 +1,82 @@
...
-    aRestricted = intersect(a, domainPow)
-    @round(T, aRestricted.lo^x, aRestricted.hi^x)
+    isinteger(x)  && return a^round(Int,x)
+    x < zero(x)  && return inv(a^(-x))
+    x == 0.5  && return sqrt(a)
+
+    zero(a.hi) > a.hi && error("Undefined: interval is strictly negative and power is non-integer")
+
+    x_interval = make_interval(T, x) # convert(Interval{T}, x)  # @interval(x) ?
+    diam(x_interval) > 2*eps(x) && return a^x_interval
```
lbenet: I think the former line was correct, since it is checking if the power is an exact representable number or not, i.e. a thin interval. If it is not, then you promote it into an interval. But if it, then the rest should yield correct results.

lbenet: I am checking what can it go wrong in travis, since the tests that don't pass are related to `^`

src/intervals/interval_functions.jl
```diff
@@ -1,95 +1,82 @@
...
+    domain = Interval{T}(0, Inf)
+    a_restricted = a ∩ domain
+
+    @round(T, a_restricted.lo^x, a_restricted.hi^x)
 end
 
 # Interval power of an interval:
 function ^{T}(a::Interval{T}, x::Interval)
-    # Is x a thin interval?
-    diam(x) < eps( mid(x) ) && return a^(x.lo)
```
lbenet: See my comment above about the case interval^thin_interval

lbenet: Sorry. I didn't notice that the lines were swapped

lbenet: .... But still the previous check was correct, i.e. `diam(x) < eps( mid(x) )`

lbenet:
I think part of the issue is related with the fact that now `@interval(-2,3)` yields an `Interval{Float64}` now, and not an `Interval{Int64}` or an `Interval{BigFloat}`. I actually have no idea how can you change this, to check any further. In any case, I think I prefer to get automatically an `Interval{Int64}` ot `Interval{BigInt}` in this case, than a Float64 interval.

But, trying to reproduce the whole problem, I noticed that
```julia
julia> @interval(-2,3)^2
[0.0, 9.000000000000002]

julia> ValidatedNumerics.@round(Float64, mig(@interval(-2,3))^2, mag(@interval(-2,3))^2)

[0.0, 9.000000000000002]

julia> ValidatedNumerics.@round(BigFloat, mig(@interval(-2,3))^2, mag(@interval(-2,3))^2)
[0.0, 9.0]
```
So, somewhere and somehow, `mag(@interval(-2,3))^2` is rounded to the next float...

dpsanders:
The rounding issue seems to just be the following:
```
julia> with_rounding(Float64, RoundUp) do
       3.0^2
       end
9.000000000000002
```

You are right that there is an issue about when to use a thin interval in the powers. However, I am worried what the effect is of using the `:wide` rounding mode instead of `:narrow`, since this *always* creates intervals of width `2 eps`...

It seems to me that the solution is to just define
`^(a::Interval, x::FloatingPoint)` and use the rule for real numbers, and separately define `^(a::Interval, x::Interval)` assuming that it is an interval`

dpsanders:
Oops, didn't mean to close.

lbenet:
After merging #26, shall I rebase or merge this branch to have an updated with respect to master?

dpsanders:
The new behaviour is as follows: 

`@interval` respects the current interval precision. Do

    set_interval_precision(100)

to set it to `BigFloat` with 100 bits, and

    set_interval_precision(Float64)

to set it back to floats.

`@interval(1, 2)` has never created an interval of `Int`s. This is done by

    Interval(1, 2)`.

Yes, it's a bit confusing...

dpsanders:
Yes, it needs to be rebased, but let's fix the problems with `ˆ` first!

dpsanders:
I have changed the `^(Interval, Real)` to `^(Interval, FloatingPoint)`. This is clearer. 
When the exponent is an interval, it still checks if the interval is "thin", defined as `<= 2eps`.
Tests now pass locally...

lbenet:
Tests are still not passing in travis....

lbenet:
> I have changed the ^(Interval, Real) to ^(Interval, FloatingPoint). This is clearer.
OK

> When the exponent is an interval, it still checks if the interval is "thin", defined as <= 2eps.
I'm not sure this is correct... The point is, do all intervals constructed from a float which is *not* exactly representable in the binary expansion have a width of 2eps? If this is true, then there is no problem. But this is *actually* not true:
```julia
julia> a = prevfloat(0.1)
0.09999999999999999
julia> @interval(a)
[0.09999999999999999, 0.1]
julia> diam(ans) == eps(0.1)
true
```

So, for `(some_interval)^a`, the conditional stated above is `true`, i.e. a will be taken as a thin interval, despite of the fact that `a.hi - a,lo > 0`...

dpsanders:
Closed in favour of #29.

Issue #23, opened 2015-05-06
Plan for v0.1


dpsanders:
### For v0.1:

**Required:**
- [x] Correct trig functions for points near boundaries of quadrants. RFC (and for tests): #20 
- [x] Documentation with `mkdocs`. RFC: #30 
- [x] Finish alternative roundings by changing tests to work with both roundings. RFC: #29
- [x] Incorporate functions from @jverzani [`ImplicitEquations` package] (https://github.com/jverzani/ImplicitEquations.jl/blob/master/src/intervals.jl) #32
- [x] Improve Krawczyk in 1D. @lbenet #26
- [x] Rename `findroots` -> `find_roots`. #26
- [x] Version of `find_roots` (with different name) returning midpoint and radius instead of interval #35 
- [x] Tests with higher precision e.g. for the Newton method #29
- [x] Add test coverage info and badge
- [x] Move rounding mode, precision, etc. into an `IntervalParameters` object #29
- [x] Add `set_interval_precision` that changes `big_pi` for interval trig #29
- [x] Or use `set_interval_precision` to choose between `Float64` and `BigFloat`. #29 
- [x] Example notebooks #36 

cc @lbenet 

**Rejected:**
- Make `@interval` default to `Float64`s and replace `@floatinterval` by `@biginterval`
[There are @floatinterval and @biginterval macros, but @interval is preferred, which now uses the `interval_parameters` object to determine the type of interval created]

- Add constructors for `Interval` that automatically round single arguments?
[I think this would be more confusing.]

dpsanders:
@jverzani Would you be happy for us to incorporate the functions you wrote?

lbenet:
Yo también trataré de avanzar en #20 y Krawczyk ;-)

jverzani:
More than. Thanks.

On Wed, May 6, 2015 at 7:19 PM, David P. Sanders <notifications@github.com>
wrote:

> @jverzani <https://github.com/jverzani> Would you be happy for us to
> incorporate the functions you wrote?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dpsanders/ValidatedNumerics.jl/issues/23#issuecomment-99641775>
> .
>



-- 
John Verzani
Chair, Department of Mathematics
College of Staten Island, CUNY
verzani@math.csi.cuny.edu

dpsanders:
@jverzani Great, thanks!

dpsanders:
[...]

PR #26, opened 2015-05-13
Rewrite Krawczyk's method, rename findroots to find_roots and add tests

lbenet:
Rewrote Krawczyk's method keeping the implementation of Newton's method
as standard. Now, both methods use the function `clean_roots!` which is
located in newton.jl.

The tests check both methods using find_roots.


lbenet:
This pull request addresses two points in the list #23, namely, Krawczyk's method and renaming `findroots`

lbenet:
Tests pass; it's ready to be merged.

lbenet:
Just pushed the new commits.

What I dislike of the idea of simply passing the operators is that it seems that it influences strongly the performance. That's the reason I tend to avoid that practice.

lbenet:
Tests pass. Incidentally, have you checked the time that the tests run in julia v0.3.8 vs v0.4-dev? It's ~1min vs 32 sec, respectively...

dpsanders:
Ah yes, you could be right about the performance when passing the operators.

dpsanders:
If you prefer, I'll just merge it and then we can work on it.

lbenet:
Let me correct the error message and see if I can do something more condensed for newton and krawczyk

lbenet:
Tests pass... I think this is ready to be merged

dpsanders:
Rebased into a single commit. (Or would you prefer to do it yourself?)

lbenet:
It's ok with me.

dpsanders:
OK, I'll merge once Travis gives the green light.

lbenet:
Green light there :-)

lbenet:
I have merged it...

dpsanders:
:+1:

PR #28, opened 2015-05-14
Lbenet/interval parameters

lbenet:
I just added a new branch, after rebasing with the current master (doing it like this, I think, respects the history), where I reverted some changes that we have discussed in #25. 

I also added new tests which involve different versions of x=1/3 as power (floating point, rational or interval). They yield consistent results. In my machine tests pass. I'd like to see if the same problem appears in travis as before.


lbenet:
Incidentally, doing things like this seems to preserve the history/authorship

lbenet:
Tests are still failing at the same place... I can't reproduce it in my machine

dpsanders:
Powers for real intervals with real exponents is known to be a complicated problem:

http://exp.ln0.de/heimlich-power-2011.htm

src/intervals/interval_functions.jl
```diff
@@ -1,95 +1,83 @@
...
 ^{T}(a::Interval{T}, r::Rational) = (a^(r.num)) ^ (1/r.den)
 
 # Real power of an interval:
-function ^{T}(a::Interval{T}, x::Real)
-    isinteger(x) && return a^round(Int,x)
-    x < zero(x) && return inv( a^(-x) )
-    x == one(x)/2 && return sqrt(a)
-    #
-    z = zero(a.hi)
-    z > a.hi && error("Undefined operation; Interval is strictly negative and power is not an integer")
```
lbenet: This is actually too simplistic; `(@interval(-1))^(1/3)` yields the error; though it should yield `Interval(-1,-1)`

src/intervals/interval_functions.jl
```diff
@@ -1,95 +1,83 @@
...
-    #
-    xInterv = convert(Interval{T}, x)  # @interval(x) ?
+function ^{T}(a::Interval{T}, x::FloatingPoint)
+    isinteger(x)  && return a^round(Int,x)
+    x < zero(x)  && return inv(a^(-x))
+    x == 0.5  && return sqrt(a)
+
+    zero(a.hi) > a.hi && error("Undefined: interval is strictly negative and power is non-integer")
```
lbenet: Ups... I meant this line

dpsanders:
Closing in favour of #29 .

PR #29, opened 2015-05-17
Interval parameters new

dpsanders:
A new version of interval parameters.

Changes the behaviour of `@interval` so it now creates intervals with the precision stored in the 
`interval_parameters` object.

Various tests had to be disabled due to apparent floating-point rounding errors on Travis.

### DIFF
--- a/src/intervals/interval_arithmetic.jl
+++ b/src/intervals/interval_arithmetic.jl
@@ -1,16 +1,25 @@
 
 ## Empty interval:
 
-emptyinterval(T::Type) = Interval(convert(T, NaN))  # interval from Inf to Inf
-emptyinterval(x::Interval) = Interval(oftype(x.lo, NaN))
+@doc doc"""Empty intervals are represented as intervals of `NaN`s.
+The automatic propagation of `NaN`s means that any operation with an empty interval gives back
+an empty interval.""" ->
+
+emptyinterval(T::Type) = Interval(convert(T, NaN))
+emptyinterval(x::Interval) = Interval(convert(eltype(x), NaN))
+
+∅ = emptyinterval(Float64)
+emptyinterval() = ∅
+
 isempty(x::Interval) = isnan(x.lo) || isnan(x.hi)
-∅ = emptyinterval(Float64)   # I don't see how to define this according to the type
 
+eps(x::Interval) = max(eps(x.lo), eps(x.hi))
 
-## "Thin" interval (no more precision):
+## "Thin" interval (one for which there is "no more precision")
+# Note that this is not the standard usage of "thin interval", which is one for
+# which the two endpoints are *strictly* equal
 
 isthin(x::Interval) = (m = mid(x); m == x.lo || m == x.hi)
-# This won't ever be the case with BigFloat if the interval is centered around 0?
 
 ## Widen:
 widen{T<:FloatingPoint}(x::Interval{T}) = Interval(prevfloat(x.lo), nextfloat(x.hi))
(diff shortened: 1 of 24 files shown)


dpsanders:
The tests are apparently finally working (having disabled various Float64 tests due to apparent errors in rounding on Travis).

I think we should merge this...

src/intervals/interval_arithmetic.jl
```diff
@@ -1,16 +1,25 @@
...
-emptyinterval(T::Type) = Interval(convert(T, NaN))  # interval from Inf to Inf
-emptyinterval(x::Interval) = Interval(oftype(x.lo, NaN))
+@doc doc"""Empty intervals are represented as intervals of `NaN`s.
+The automatic propagation of `NaN`s means that any operation with an empty interval gives back
+an empty interval.""" ->
+
+emptyinterval(T::Type) = Interval(convert(T, NaN))
+emptyinterval(x::Interval) = Interval(convert(eltype(x), NaN))
+
+∅ = emptyinterval(Float64)
```
lbenet: I think it is a good idea to define `∅` as a `const` which depends of the type of intervals that are being used. Something like the way `pi` is treated

src/intervals/interval_arithmetic.jl
```diff
@@ -44,7 +54,7 @@ one{T}(a::Interval{T}) = Interval(one(T))
 
 -{T}(a::Interval{T}) = @round(T, -a.hi, -a.lo)
 -(a::Interval, b::Interval) = a + (-b)  # @round(a.lo - b.hi, a.hi - b.lo)
-
+(((())))
```
lbenet: ¿?

dpsanders: No idea, sorry!

src/intervals/interval_arithmetic.jl
dpsanders: Good idea: #33

src/intervals/interval_arithmetic.jl
dpsanders: I have removed this in a later branch. For the moment it does no harm...

src/intervals/interval_functions.jl
```diff
@@ -1,95 +1,105 @@
...
-    #
-    z = zero(a.hi)
-    z > a.hi && error("Undefined operation; Interval is strictly negative and power is not an integer")
-    #
-    xInterv = convert(Interval{T}, x)  # @interval(x) ?
-    diam( xInterv ) >= eps(x) && return a^xInterv
+function ^{T}(a::Interval{T}, x::FloatingPoint)
+    isinteger(x)  && return a^(round(Int,x))
+    x < zero(x)  && return inv(a^(-x))
+    x == 0.5  && return sqrt(a)
```
lbenet: This relates to our long-lasting discussion on conversion... I think that instead of
`x == 0.5 ` it should be written `x == T(0.5)` or something alike, to avoid conversion.
I guess that this imposes some use of `@compat`, since it is a v0.4 usage.

src/intervals/interval_functions.jl
```diff
@@ -1,95 +1,105 @@
...
            end,
            begin
-               hilo = aRestricted.hi^(x.lo)
-               hihi = aRestricted.hi^(x.hi)
+               hilo = a_restricted.hi^(x.lo)
+               hihi = a_restricted.hi^(x.hi)
                max( hilo, hihi)
            end
            )
 end
```
lbenet: Nice!

src/root_finding/newton.jl
```diff
@@ -1,51 +1,62 @@
...
+
+    m - (f(m) / deriv)
 end
 
 
+@doc doc"""If a root is known to be inside an interval, `newton_refine`
+just iterates the interval Newton method until that root is found.""" ->
 function newton_refine{T}(f::Function, f_prime::Function, x::Interval{T};
-    tolerance=eps(one(T)), debug=false)
+                          tolerance = 10*eps(T), debug = false)
```
lbenet: Why such a large tolerance? It used to work with `eps(T)`...

lbenet: Also, this value of tolerance is changed in `newton`!?

lbenet:
Could you rebase to clean up what is already merged? Some of the commits included here are already merged

dpsanders:
I couldn't get the rebase to work, so I just merged it as is (green light on Travis).

PR #32, opened 2015-05-18
Incorporate some suggestions by John Verzani and more clean-up

dpsanders:


dpsanders:
Incorporate some suggestions by @jverzani from https://github.com/jverzani/ImplicitEquations.jl/blob/master/src/intervals.jl:

- inverse trig functions (`asin`, `acos`, `atan`) -- the implementation restricts the domain to `[-1,1]` for `asin` and `acos`
- `floor` and `ceil`

It is not clear to me that `<=`, `max` and `min` make sense in the general context of intervals, so I have not included them. Do you need them, @jverzani?

jverzani:
I use <= and others, but they make more sense in my context where the answer is not boolean, but also has a possible "MAYBE" state. I'd leave out of your package if they don't make sense. Thanks!

lbenet:
This is ready to be merged, indeed, though better after #29

dpsanders:
Rebased and green. Merging.

PR #35, opened 2015-05-22
Adds functions to calculate the midpoint_radius representation.

dpsanders:
Adds a version of find_roots that returns a tuple of the midpoints, a tuple of the radii, and a tuple of the symbols

### DIFF
--- a/src/root_finding/root_finding.jl
+++ b/src/root_finding/root_finding.jl
@@ -1,39 +1,127 @@
 
+# immutable Root{T<:Real}
+#     interval::Interval{T}
+#     existence::Symbol    # :unique or :unknown
+# end
+
+
+
+# Base.getindex(r::Root, i::Integer) = getfield(r, i)
+
+@which sort!([3, 5, 4])
+
+#typealias Root{T<:Real} @compat Tuple{Interval{T}, Symbol}
+immutable Root{T<:Real}
+    interval::Interval{T}
+    root_type::Symbol
+end
+
+show(io::IO, root::Root) = print(io, "Root($(root.interval), :$(root.root_type))")
+
+is_unique{T}(root::Root{T}) = root.root_type == :unique
+
+
 include("automatic_differentiation.jl")
 include("newton.jl")
 include("krawczyk.jl")
 
 
-typealias Root{T<:Real} @compat Tuple{Interval{T}, Symbol}
+
 
 function find_roots{T}(f::Function, a::Interval{T}, method::Function = newton;
                     tolerance = eps(T), debug = false, maxlevel = 30)
 
-    method(f, a, tolerance=tolerance, debug=debug, maxlevel=maxlevel)
+    method(f, a; tolerance=tolerance, debug=debug, maxlevel=maxlevel)
 end
 
 function find_roots{T}(f::Function, f_prime::Function, a::Interval{T}, method::Function=newton;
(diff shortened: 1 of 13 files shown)


dpsanders:
Well let's just leave it as `find_roots_midpoint` for the moment and think about a better interface.
This is ready to be merged; the red X comes from coveralls and not from Travis. (I.e., we need more tests...)

dpsanders:
Squashed. Ready to merge I think.

lbenet:
@dpsanders I really like lots of the changes and additions. 

One suggestion: Change the output of `midpoint_radius`. Currently we get:
```julia
julia> @interval 0.1
[0.09999999999999999, 0.1]

julia> midpoint_radius(ans)
(0.1,1.3877787807814457e-17)
```

I think it looks nicer that the output of the last line is something like: 
```
0.1 ± 1.3877787807814457e-17
```

lbenet:
Along the same lines:
```julia
julia> set_interval_precision(BigFloat, 53)
53

julia> @interval 0.1
[9.9999999999999992e-02, 1.0000000000000001e-01]₅₃

julia> midpoint_radius(ans)
(1.0000000000000001e-01,1.3877787807814457e-17)
```
The last line doesn't show that it is a BigFloat interval of 53 bits of precision...

src/root_finding/root_finding.jl
```diff
@@ -1,39 +1,127 @@
 
+# immutable Root{T<:Real}
```
lbenet: I really like the idea; have you benchmark it against previous behavior? I would guess it is much faster now, isn't it?

lbenet: This was meant on line 14...

lbenet:
I'll wait for your reply about my [comment](https://github.com/dpsanders/ValidatedNumerics.jl/pull/35#issuecomment-113973101). I agree it is ready to be merged.

src/root_finding/root_finding.jl
dpsanders: I haven't benchmarked it. The change was necessary to have more readable code (and results). I expect it doesn't change the speed too much (at least on v0.4).

dpsanders:
Your two comments about the display are good ones, but not relevant for this PR:

1. The first comment, about output like `3 ± 0.1`, seems to me to be really about how intervals should be displayed. We could add a global option to change this output for displaying any intervals. 

The function `midpoint_radius` is designed only to calculate and return the values for later use; changing its output would require actually making a new type with its own `show`, but really it's just a different representation of the same interval, so this doesn't seem like a good idea. 

I have opened issue #39 for this, but I think we can leave this til after 0.1.

2. The issue about how `BigFloat`s are displayed is due to the recent change https://github.com/JuliaLang/julia/pull/11578 in base Julia; in v0.3 the output is

```julia
(1.0000000000000001e-01 with 53 bits of precision,5.551115123125783e-18 with 53 bits of precision)
```

We can fix this in #39 for intervals; this example with 53 bits suggests that the idea of just removing the indication of the precision in the output of `BigFloat`s was not necessarily a good idea...!

dpsanders:
The red dot is due to the coverage decrease; tests pass on Travis. 
We definitely need better test coverage, but merging for the moment.

lbenet:
I agree with both comments, and that this is merged.

PR #37, opened 2015-05-24
Add MathJax support to the documentation and the repo address

lbenet:
I also included my web page in README.md and included travis and coveralls
badges. Incidentally, the latter was pointing to interval_parameters branch;
the current points to the master branch

In docs/root_finding.md I had to use \ast instead of * to make MathJax
understanding that the asterix was not meant as italics but as an asterix.


lbenet:
@dpsanders I have no idea how to update this changes to the documentation; could you please do that?

PR #38, opened 2015-05-24
Notebooks with examples of use of the package

dpsanders:


dpsanders:
Now has just the current versions of notebooks.

lbenet:
Very nice! I'm merging it.

PR #54, opened 2015-09-01
Adds zero(::Type{Interval}) and one(::Type{Interval}), and tests

lbenet:
The reason to add these methods is to keep up with the usage of zero/one
in Base, and because it is useful to play smoothly with TaylorSeries.
It also deletes two methods of ⊆, which are already in the package,
which were left in root_finding_tests.jl

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -44,7 +44,9 @@ in(x::Real, a::Interval) = a.lo <= x <= a.hi
 ## zero and one functions
 
 zero{T}(a::Interval{T}) = Interval(zero(T))
+zero{T}(::Type{Interval{T}}) = Interval(zero(T))
 one{T}(a::Interval{T}) = Interval(one(T))
+one{T}(::Type{Interval{T}}) = Interval(one(T))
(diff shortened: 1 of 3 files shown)


lbenet:
This is a very tiny and minor PR. Yet, it allows to use `ValidatedNumerics` and `TaylorSeries` together in a very smooth way. In particular, I can then define `Taylor1{Interval{T}}`, which is quite important for implementing validated higher integration methods.

dpsanders:
As I suspected, there is already a generic fallback (before this branch):

```julia
julia> one(Interval{Float64})
[1.0, 1.0]

julia> @which one(Interval{Float64)
ERROR: syntax: missing comma or } in argument list

julia> @which one(Interval{Float64})
one{T<:Number}(::Type{T<:Number}) at number.jl:51
```

Nonetheless, it certainly doesn't hurt to add these explicitly.

lbenet:
Mmmm. It seems that I require this to play smoothly in TaylorSeries. I'll send you an exmaple later.

lbenet:
Thanks for merging

PR #55, opened 2015-09-09
Towards conforming IEEE-1788 standard

lbenet:
This branch includes some changes I've implemented to eventually conform with the IEEE-1788 standard. I'm trying to go step by step as I am reading the simplified version of the standard. 

There are many issues to be solved yet.

cc @oheim

### DIFF
--- a/src/ValidatedNumerics.jl
+++ b/src/ValidatedNumerics.jl
@@ -29,7 +29,7 @@ export
     diam, mid, mag, mig, hull, isinside,
     emptyinterval, ∅, isempty, ⊊,
     entireinterval, isentire, nai, isnai,
-    widen,
+    widen, infimum, supremum,
     set_interval_precision, get_interval_precision,
     with_interval_precision,
     interval_parameters, eps, dist, roughly,
--- a/src/ValidatedNumerics.jl
+++ b/src/ValidatedNumerics.jl
@@ -12,7 +12,7 @@ using Compat
 
 import Base:
     +, -, *, /, //,
-    <, >, ==, !=, ⊆, ^, ⊊,
+    <, >, ==, !=, ⊆, ^, <=,
     in, zero, one, abs, real, show,
     sqrt, exp, log, sin, cos, tan, inv,
     asin, acos, atan,
@@ -28,9 +28,10 @@ export
     Interval,
     @interval, @biginterval, @floatinterval, @make_interval,
     get_interval_rounding, set_interval_rounding,
-    diam, mid, mag, mig, hull, isinside,
+    diam, radius, mid, mag, mig, hull, isinside,
     emptyinterval, ∅, isempty, interior, isdisjoint, ⪽,
-    entireinterval, isentire, nai, isnai, isthin,
+    precedes, strictprecedes, ≺,
+    entireinterval, isentire, nai, isnai, isthin, iscommon,
     widen, infimum, supremum,
     set_interval_precision, get_interval_precision,
     with_interval_precision,
--- a/src/ValidatedNumerics.jl
+++ b/src/ValidatedNumerics.jl
@@ -28,7 +28,7 @@ export
     get_interval_rounding, set_interval_rounding,
     diam, mid, mag, mig, hull, isinside,
     emptyinterval, ∅, isempty, ⊊,
-    entireinterval, isentire,
+    entireinterval, isentire, nai, isnai,
     widen,
     set_interval_precision, get_interval_precision,
     with_interval_precision,
--- a/src/ValidatedNumerics.jl
+++ b/src/ValidatedNumerics.jl
@@ -10,7 +10,7 @@ using Compat
 
 import Base:
     +, -, *, /, //,
-    <, >, ==, !=, ⊊, ⊆, ^,
+    <, >, ==, !=, ⊆, ^, ⊊,
     in, zero, one, abs, real, show,
     sqrt, exp, log, sin, cos, tan, inv,
     asin, acos, atan,
@@ -27,8 +27,8 @@ export
     @interval, @biginterval, @floatinterval, @make_interval,
     get_interval_rounding, set_interval_rounding,
     diam, mid, mag, mig, hull, isinside,
-    emptyinterval, ∅, isempty, ⊊,
-    entireinterval, isentire, nai, isnai,
+    emptyinterval, ∅, isempty, interior, isdisjoint, ⪽,
+    entireinterval, isentire, nai, isnai, isthin, 
     widen, infimum, supremum,
     set_interval_precision, get_interval_precision,
     with_interval_precision,
--- a/src/ValidatedNumerics.jl
+++ b/src/ValidatedNumerics.jl
@@ -28,6 +28,7 @@ export
     get_interval_rounding, set_interval_rounding,
     diam, mid, mag, mig, hull, isinside,
     emptyinterval, ∅, isempty, ⊊,
+    entireinterval, isentire,
     widen,
     set_interval_precision, get_interval_precision,
     with_interval_precision,
--- a/src/ValidatedNumerics.jl
+++ b/src/ValidatedNumerics.jl
@@ -1,3 +1,5 @@
+# This file is part of the ValidatedNumerics.jl package; MIT licensed
+
 VERSION >= v"0.4.0-dev+6521" && __precompile__(true)
 
 module ValidatedNumerics
@@ -28,7 +30,7 @@ export
     get_interval_rounding, set_interval_rounding,
     diam, mid, mag, mig, hull, isinside,
     emptyinterval, ∅, isempty, interior, isdisjoint, ⪽,
-    entireinterval, isentire, nai, isnai, isthin, 
+    entireinterval, isentire, nai, isnai, isthin,
     widen, infimum, supremum,
     set_interval_precision, get_interval_precision,
     with_interval_precision,
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -15,6 +15,7 @@ emptyinterval() = ∅
 # isempty(x::Interval) = isnan(x.lo) || isnan(x.hi)
 isempty(x::Interval) = x.lo == Inf && x.hi == -Inf
 
+
 @doc doc"""`entireinterval`s represent the whole Real line: [-∞, ∞].""" ->
 entireinterval(T::Type) = Interval(convert(T, -Inf), convert(T, Inf))
 entireinterval(x::Interval) =
@@ -22,11 +23,13 @@ entireinterval(x::Interval) =
 
 isentire(x::Interval) = x.lo == -Inf && x.hi == Inf
 
-# Do we need the following? Seems to be more julian...
-# inf{T<:Real}(::Type{Interval{T}}) = entireinterval(T::Type)
-# inf{T<:Real}(x::Interval) = entireinterval(x)
-#
-#isinf(x::Interval) = isentire(x)
+
+@doc doc"""`NaI` not-an-interval: [NaN, NaN].""" ->
+nai(T::Type) = Interval(convert(T, NaN), convert(T, NaN))
+nai(x::Interval) = Interval(convert(eltype(x), NaN), convert(eltype(x), NaN))
+
+isnai(x::Interval) = isnan(x.lo) || isnan(x.hi)
+
 
 eps(x::Interval) = max(eps(x.lo), eps(x.hi))
 
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -57,11 +57,11 @@ end
 
 ## Scalar functions on intervals (no directed rounding used)
 
-mid(a::Interval) = (a.lo + a.hi) / 2
+mid(a::Interval) = ifelse(isentire(a), zero(a.lo), (a.lo + a.hi) / 2)
 
 diam(a::Interval) = a.hi - a.lo
 mag(a::Interval) = max( abs(a.lo), abs(a.hi) )
-mig(a::Interval) = ( zero(a.lo) ∈ a ) ? zero(a.lo) : min( abs(a.lo), abs(a.hi) )
+mig(a::Interval) = ifelse(zero(a.lo) ∈ a, zero(a.lo), min(abs(a.lo), abs(a.hi)))
 
 
 # Infimum and supremum of an interval
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -10,13 +10,17 @@
 
 ## Inclusion/containment functions
 
-in{T<:Real}(x::T, a::Interval) = a.lo <= x <= a.hi
+function in{T<:Real}(x::T, a::Interval)
+    isentire(a) && return true
+    a.lo <= x <= a.hi
+end
 
 # issubset
 ⊆(a::Interval, b::Interval) = b.lo ≤ a.lo && a.hi ≤ b.hi
 
 function interior(a::Interval, b::Interval)
     (isempty(a) && isempty(b)) && return true
+    (isentire(a) && isentire(b)) && return true
     b.lo < a.lo && a.hi < b.hi
 end
 const ⪽ = interior  # \subsetdot
@@ -57,11 +61,25 @@ end
 function *{T<:Real}(a::Interval{T}, b::Interval{T})
     (isempty(a) || isempty(b)) && return emptyinterval(T)
 
-    (zero(a) == a || zero(b) == b) && return zero(Interval{T})
+    (a == zero(a) || b == zero(b)) && return zero(a)
 
-    @round(T, min( a.lo*b.lo, a.lo*b.hi, a.hi*b.lo, a.hi*b.hi ),
-            max( a.lo*b.lo, a.lo*b.hi, a.hi*b.lo, a.hi*b.hi ) )
+    if b.lo >= zero(T)
+        a.lo >= zero(T) && return @round(T, a.lo*b.lo, a.hi*b.hi)
+        a.hi <= zero(T) && return @round(T, a.lo*b.hi, a.hi*b.lo)
+        return @round(T, a.lo*b.hi, a.hi*b.hi) # in(zero(T), a)
+    elseif b.hi <= zero(T)
+        a.lo >= zero(T) && return @round(T, a.hi*b.lo, a.lo*b.hi)
+        a.hi <= zero(T) && return @round(T, a.hi*b.hi, a.lo*b.lo)
+        return @round(T, a.hi*b.lo, a.lo*b.lo) # in(zero(T), a)
+    else
+        a.lo > zero(T) && return @round(T, a.hi*b.lo, a.hi*b.hi)
+        a.hi < zero(T) && return @round(T, a.lo*b.hi, a.lo*b.lo)
+        return @round(T, min(a.lo*b.hi, a.hi*b.lo), max(a.lo*b.lo, a.hi*b.hi))
+    end
+    # @round(T, min( a.lo*b.lo, a.lo*b.hi, a.hi*b.lo, a.hi*b.hi ),
+    #         max( a.lo*b.lo, a.lo*b.hi, a.hi*b.lo, a.hi*b.hi ) )
 end
+
 
 ## Division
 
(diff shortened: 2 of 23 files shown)


oheim:
This definitely goes in the right direction!

Some remarks that come to my mind (please double check, I am writing this w/o looking at the standard document):
* You may encode the empty set as either (nan, nan) or (inf, -inf) or whatever else you prefer if it helps interval operations to behave correctly. The first is sometimes beneficial in arithmetic operations. The latter is sometimes beneficial in boolean operations. However, inf(empty) (or empty.lo in your case) must return positive infinity.
* NaI does only exist in decorated interval arithmetic, there is no undecorated (a.k.a. bare) NaI. You could alternatively use (nan, nan) = empty; (nan, nan, trv) = decorated empty; and (nan, nan, ill) = NaI.
* Illegal interval construction may throw an exception in bare interval arithmetic and must return NaI in decorated interval arithmetic. For example `@interval(3, 2)` => error, and `@decoratedinterval(3, 2)` => nai – maybe other behaviour is also allowed, but that's how I have implemented it in Octave ;-)

oheim:
P.S. Decorated interval arithmetic is a great feature and I recommend to make it easy to use—easier than undecorated interval arithmetic. Currently you have the `@interval` macro and the `Interval` constructor. As soon as you support decorated interval arithmetic you will have (at least) two constructors: for bare and decorated intervals. I suggest that you keep the `@interval` macro, but make it produce a decorated interval by default. Then, users attention will be directed to that great feature.

lbenet:
Thanks for the comments and suggestions, Olivier. 

I/We do not have the standard yet, so what I did yesterday was based in the few pages I have read from the simplified version, some intuition, and what I thought we were not following.

Some comments and thoughts:

- I think that we should concentrate (for the time being) in making the bare intervals compliant with the standard; the extension to decorated intervals is simple (once we decide about the actual form, see below).

- I thought the standard was imposing that the empty set was defined as `[Inf,-Inf]`; see in Sect. 3.2 the definition of bounds. (Before we had defined it with `NaN`'s, this is currently `nai`.) The present form should make some of the test that previously failed to pass.

- Regarding `nai`'s for undecorated intervals, I agree with your comment that they should be implemented for decorated intervals. The question is, what should it be the outcome of `[0,0] / [0,0]` for undecorated intervals according to the standard ?

- Regarding illegal interval construction, say [1,0], our current implementation currently translates such cases (without any warning) into a properly defined interval, e.g. [0,1]. Must we really throw an error (or a warning) in those cases? The exceptions to the former is the construction of `emptyinterval` (or `nai`) by using the `Interval` constructor; if you use `@interval` you get that the lower bound is less than the upper one. 

- I think we need to think the way we want to implement decorated intervals. We can simply extend the current definition, with a new field `dec`  for the type `Interval`, which maybe one of the 5 decorations mentioned in the standard, or `#undef` for undecorated intervals. Another possibility is to duplicate essentially everything with a new type `DecInterval`, which has that field. I haven't yet thought what type of field should `dec` be. Is the decoration field fixed throughout the calculation? If yes, we could define it by an integer (from 0 to 5), and display it nicer to the user, keeping `Interval` as an immutable; if not, maybe a one-component integer vector could be used. What do you think @dpsanders ?

oheim:
Sounds like a good plan to first make the bare interval arithmetic conforming and then move on to decorated intervals. Some recommendations regarding your questions:

1. The lower and upper bound of the empty interval are defined as `+Inf` and `-Inf` respectively. However, the users accesses these through the inf and sup function, so you *could* use `NaN` for internal representation if you wish.
2. The outcome of `[0,0] / [0,0]` is the empty set. These are semantics of “set-based” interval arithmetic. See 4.4.4 for the definition of an interval extension of a point function. The input values where the function is undefined are ignored. Also for example you have `log ([-2, 0]) = [empty]`. There will a difference in decorated interval arithmetic: The result decoration will be `trv` if the function has been evaluated outside of its domain.
3. The constructor `numsToInterval` as defined by the standard must not return a value if infimum is greater than supremum— as must the constructor `textToInterval` on illegal interval literals. Again, legal interval literals (see 6.6.2) must not have infimum greater than supremum. Even `[+Inf, -Inf]` is illegal if you strictly follow the standard.  
It is not quite clear what “no value” means depending on the implementation, but I would translate it with raising an exception.
4. Implementing the decorated intervals should be well thought of. There has been much discussion on the IEEE 1788 mailing list already. For example you have to take care that the user cannot accidentally mix decorated intervals during computation. In some languages it would be dangerous to have the decorated interval type as an extension of the bare interval type. E.g. [1,2] + [2,3]_com is dangerous, because one interval is missing the decoration and you cannot safely assume a non-trivial decoration for the final result. On the other side the decorated variant of textToInterval will create [1,2]_com from the interval literal [1,2], so it is not trivial to understand.  
In Octave I have allowed arithmetic of the form [1,2] + [2,3]_com but the library throws a warning. The user has the choice to ignore these warnings or handle them like exceptions, whatever she prefers.
5. A natural choice for internal encoding of the decorations would be a uint8 with the interchange encoding defined in section 7.3. However, feel free to use an enumeration or whatever Julia has to offer.

lbenet:
Tests are not passing, probably due to the fact that there is a version 0.4-rc1...

lbenet:
My mistake; I had included a test which was not correct. Now tests pass.

oheim:
@lbenet, I have updated the plugin for ValidatedNumerics.jl in ITF1788 to reflect your recent changes in this branch. Please also note how you can define alternative syntaxes for particular functions. The generated tests will then check both.

You should maintain that plugin yourself as you progress with your library. I have granted you push access to my fork of ITF1788, so you don't have to fork yourself. Please ask if you would like to know how to do certain things with that generator.

dpsanders:
Many thanks for your comments and excellent test package, @oheim!

Nice work, @lbenet! 
I have made a few comments about style. Once these are addressed, maybe we can merge this and carry on with a new PR. The more difficult points can be addressed separately.

lbenet:
@oheim Thanks for updating the plugin for julia; this afternoon I'll be traveling, so I will take a look how things work. Any guidance will be appreciated. Incidentally, I have an issue (some weird error that seems related to julia, but that I can't really track) which prevents me from rerunning the tests; can you explain me how can I do this?

Thanks also for your comments on all subtleties regarding the standard. I'll try to continue adapting what we now have to the standard.

@dpsanders Thanks for reviewing the code; I'll try to work on the observations this afternoon while traveling. The conference ends at noon today.

oheim:
@dpsanders, the test package is not my own work. Planing and most test cases come from @nehmeier, python programming has been done by @kiesner. I did collect the test cases and converted them into the portable itl format (which was dull work, but very useful), did some testing and advertising.

@lbenet, I am not sure what you mean, but if you use the Makefile for test execution they won't be reexecuted unless you delete the log file or regenerate the jl test scripts. So if you change ValidatedNumerics.jl and then want to reexecute all tests, that's not possible unless you change the Makefile. It's probably easier to use a test script that executes all jl test files in order.

Also please not that you want to call the itf1788 generator with the option `-c "(julia, *, *)"` otherwise you would also generate test scripts for all other plugins, which takes significantly longer.

lbenet:
@oheim The ITF1788 is really **REALLY** great! I somehow solved the problems to allow rerunning the scripts. Thanks!

Regarding my last commit, in my machine tests pass correctly, but something happens in travis that they don't (the two tests that don't pass are related with `tan`). I'll concentrate on the ITF1788 tests, since there are still a bunch that are not yet passing.

oheim:
@lbenet, I have looked at the two failing tests. The expected and actual results are what would also be returned by the octave interval package. Either both implementations behave equally wrong or the test cases are wrong. Where do the two test cases come from? Have the test cases been verified already?

````
cos(@interval(1.67,3.2)) --> Interval(-1,-0.0990410365987278)
tan(@interval(1.67,3.2)) --> Interval(-10.047182299210329,0.05847385445957865)
````

**Comment by lbenet** [comment]
@oheim I don't quite remember where do they come from; I'll check this later.

The problem I'd like to solve now is the fact that we get wider intervals in cases we shouldn't (for  `Float64`s which are have an exactly representable binary expansion), e.g., `@interval(2.5^3)`. I think this is at the heart of this kind of problems.

At a very early stage, I remember we were creating them using `BigFloat`'s with extended precision with correct rounding, but it was too slow. @dpsanders, maybe you have some ideas for this. As a concrete example, `@interval(2.5^3)` should yield the thin interval `Interval(15.625,15.625)`, but it currently yields `[15.624999999999998, 15.625000000000002]`...

**Comment by dpsanders** [comment]
> The problem I'd like to solve now is the fact that we get wider intervals in cases we shouldn't (for Float64s which are have an exactly representable binary expansion), e.g., @interval(2.5^3). I think this is at the heart of this kind of problems.

> At a very early stage, I remember we were creating them using BigFloat's with extended precision with correct rounding, but it was too slow. @dpsanders, maybe you have some ideas for this. As a concrete example, @interval(2.5^3) should yield the thin interval Interval(15.625,15.625), but it currently yields [15.624999999999998, 15.625000000000002]...

I believe this is due to the fact that we are using the built-in Julia `^` function for `Float64`s, which is *not correctly rounded*. Using `BigFloat`s instead gives the correct answer. There does not seem to be any simple solution to this problem, except to convert to `BigFloat` and back for these functions.

For some functions, e.g. `sin`, `exp` etc., this will be solved by wrapping the `crlibm` library. For others, e.g. `^`, the `crlibm` library does not contain correctly-rounded versions, so indeed the *only* solution I can see is to do the round-trip via `BigFloat`...

**Comment by dpsanders** [comment]
This can be achieved (for the moment) by something like

```julia
sin(x::Interval{Float64}) = @floatinterval(@biginterval(x))
```

dpsanders:
From what I understood that Warwick Tucker said, the software that he has been involved in, [Campary](http://homepages.laas.fr/mmjoldes/campary/), is another option.

oheim:
A correctly rounded power function is difficult to implement and fortunately MPFR has one. Maybe in Julia the power function is computed in two steps as exp (y * ln (x)). In my [thesis](http://exp.ln0.de/heimlich-power-2011.htm) I have reasoned a worst-case relative error boundary of 2^-41 for that formula, that is 2^12 ULPs in binary64 precision.

I don't know how Julia handles the built-in functions with directed rounding. In C++ the rounding modes affect several mathematical functions (libm), but they do not guarantee any accuracy or to be correctly rounded. Everything but the very basic functions (add sub div mul sqrt) is platform dependent. I have observed cases where rounding mode switches either (1) did not affect the result, (2) affected the result chaotically (one rounding mode produced a value < 1, another rounding mode resulted in > 20000), or (3) affected the result somehow, but the function's inaccuracy was too low, that is, downward rounded result < upward rounded result < mathematically exact value.

I don't know what your performance and accuracy expectations are. For the Octave interval package I have created [some benchmark numbers for the correctly rounded power function used in interval arithmetic](http://octave.sourceforge.net/interval/package_doc/Brief-Introduction.html#What-to-Expect). You could either make a fast interval power function (which still contains the correct answer) or a correctly rounded power function.

The ITF1788 framework supports test cases which reflect this dilemma. You can write
```javascript
pown [2.5,2.5] 3 = [15.625,15.625] < [15.624999999999998, 15.625000000000002];
```
Then the test case will pass even for the less accurate result. It is possible to produce warnings if the full accuracy is not reached (cf. FIXME in `test/FactCheck/test.yaml`).

lbenet:
> I believe this is due to the fact that we are using the built-in Julia `^` function for `Float64`s, which is not correctly rounded. Using `BigFloat`s instead gives the correct answer. There does not seem to be any simple solution to this problem, except to convert to `BigFloat` and back for these functions.

I'm not quite sure that the problem is the built-in `^`: `2.5^3` yields `15.625` as it should (`RoundNearest`); however, what `@interval(2.5^3)` returns is equivalent to `[prevfloat(15.625), nextfloat(15.625)]`, mimicking what `:wider` does, despite of the fact that we are in `:narrow` mode...

dpsanders:
The problem is the *rounded* version of `^`:

```julia
julia> a = with_rounding(Float64, RoundUp) do
       2.5^3
       end
15.625000000000002
```

lbenet:
Good catch, @dpsanders!

Either we get back to our old way of constructing intervals, through `BigFloat`s with correct rounding, or we should be more subtle in `@round` and `@thin_round` when dealing with `Floats64`s.

dpsanders:
I think the solution for the moment is what I suggested above: for `^` and for `sin` etc, we do the explicit conversion to and from `BigFloat`. There is no need to change anything for `+`, `-`, `*`, `/` or `sqrt`, since they are guaranteed to be correctly rounded.

lbenet:
I agree. A possible implementation could go along the following:

```julia
julia> @biginterval(2.5^3)  # I'm using bigfloat precision 53, to cut down output
[1.5625000000000000e+01, 1.5625000000000000e+01]₅₃

julia> float(ans)  # for conversion to Interval{Float64}
[15.625, 15.625]
```
The final result is what we want.

So it is sufficient to use `@biginterval` constructor, and the converting it to a float interval with `float`. Where should this be done?

dpsanders:
This can be done as I suggested above with `sin`: we literally just add the new methods as
```julia
^(x::Interval{Float64}, n::Integer) = float(@biginterval(x)^n)

sin(x::Interval{Float64}) = float(@biginterval(x))
```
etc.
(Sorry, forgot the `sin` in the previous comment.)  

Hmmm, I think rather we should in fact add a special `big53` function that converts an `Interval{Float64}` to an `Interval{BigFloat}` with the special 53-bit precision. 

Since these are more specialised than with `Interval{T}`, they will be used in preference.

dpsanders:
Here it is:
```julia
function big53(x::Interval{
julia> function big53(x::Interval{Float64})
       with_interval_precision(53) do
       @biginterval(x)
       end
       end
big53 (generic function with 1 method)

julia> x = @interval(0.1)
[0.09999999999999999, 0.1]

julia> big53(x)
[9.9999999999999992e-02, 1.0000000000000001e-01]₅₃
```

dpsanders:
(Though there is probably a more efficient way of implementing it)

lbenet:
What I meant, is whether this could be implemented at the level of macros, which is something nice and attractive now. 

I'm thinking on something like `@interval(2.5^3)` yielding `float(@biginterval(2.5^3))` (instead of the current result), and `(Interval(2.5,2.5))^3` being also equivalent to `float((Interval{BigFloat}(2.5,2.5))^3)`.

dpsanders:
> I'm thinking on something like @interval(2.5^3) yielding float(@biginterval(2.5^3)) (instead of the current result), and (Interval(2.5,2.5))^3 being also equivalent to float((Interval{BigFloat}(2.5,2.5))^3).

Putting together my above suggestions makes this automatic with no further work, as below.

I don't really want to mess with the macros as you suggest, since we don't want/need to wrap the literals (2.5 here) in `@biginterval` to do operations like `+`. I think we just use multiple dispatch. In any case, once we wrap `crlibm` (maybe now is a good time), we will, I believe, (unfortunately) need different versions of `sin` for intervals with `Float64` or `BigFloat`.

 ```
julia> function big53(x::Interval{Float64})
                  with_interval_precision(53) do
                      @biginterval(x)
                  end
              end
big53 (generic function with 1 method)

julia> import Base.^

julia> ^(x::Interval{Float64}, n::Integer) = float(big53(x)^n)
^ (generic function with 52 methods)

julia> @interval 2.5^3
[15.625, 15.625]
```

**Comment by dpsanders** [comment]
I'll make a PR with these changes!

**Comment by dpsanders** [comment]
Another example:

```
julia> @interval exp(1)
[2.718281828459045, 2.718281828459045]   # currently not correctly rounded

julia> Base.exp(x::Interval{Float64}) = float(exp(big53(x)))
exp (generic function with 15 methods)

julia> @interval exp(1)
[2.718281828459045, 2.7182818284590455]   # now *is* correctly rounded
```

**Comment by dpsanders** [comment]
Of course, this will change the results of a *lot* of tests... But it should make the ITF1788 tests run correctly! (Although maybe not for the trig functions, where we have a funky implementation...)

**Comment by lbenet** [comment]
Sorry, I missed [this](https://github.com/dpsanders/ValidatedNumerics.jl/pull/55#issuecomment-139880755) comment. I actually like the idea of `big53(x)`; I simply was wondering if it could be implemented at the level of macros, because this simplifies life later, specially at the level of the functions. :-)

**Comment by dpsanders** [comment]
In my opinion, the macros are already hairy enough! ;) 
[OT: How do you get a link to a specific comment?]

**Comment by dpsanders** [comment]
Re. the current PR: I suggest that for the moment you change the 2 failing tests so they pass and we merge ...

**Comment by lbenet** [comment]
> [OT: How do you get a link to a specific comment?]  

You copy-paste the link that appears in the time referring to the specific comment

> Re. the current PR: I suggest that for the moment you change the 2 failing tests so they pass and we merge ...

Regarding this PR, I think it is worth using it to restrict the field of numbers that define `Intervals`, to `IReal <: Real`, instead of `Real` as it is now. `IReal` will correspond to everything but `Integer`s. I think this can save us some type-instability issues (e.g., currently present in `^`, which for `Interval{Int64}` may return an `Interval{Int64}` or a `Interval{Float64}`).

**Comment by lbenet** [comment]
@dpsanders Some of the functions that have been already implemented, are slightly different from those at the standard. I will update them, hoping that this has no further consequences with respect to the tests.

I spent some time today checking trying to restrict `Interval{T}` to `T<:IReal`; this had some negative consequences related to promotion and conversion; for the time being I will leave it as it is now. 

Shall I correct the following (odd) behavior?
```julia
julia> Interval(pi)
[π = 3.1415926535897..., π = 3.1415926535897...]
```
I think this should behave as what `@floatinterval(pi)`, or `@biginterval(pi)` do, depending on `interval_parameters.precision_type`. What do you think?

lbenet:
I pushed another commit which deals with the correct behavior of a bunch of functions defined in the standard. With the current commit, all currently available tests (I added new yesterday) of the [ITF1788 suite](https://github.com/oheim/ITF1788) *pass*, except for those related to the inaccuracies in the elementary functions implementations.

I will check if I can implement (without too much changes) the restriction of `Interval` to `AbstractFloat`, `BigFloat` or `Rational`, which I proposed earlier, and also an easy work around of the current constructors to avoid the creation of too wide intervals for `Float64`.

dpsanders:
I think we should merge this PR as it is.

dpsanders:
I think we should not modify the bare `Interval` constructor. That is why `@interval` exists.

dpsanders:
Looks great, @lbenet, thanks!

lbenet:
What I thought we could do with the `Interval` is, whenever we use the constructor with `Integer` or `Irrational`, simply convert it to the current `interval_parameters.precision_type`. This can be done easily as an external constructor, so the actual `Interval` type is not modified. For example,
```julia
julia> Interval(pi)
[3.141592653589793, 3.141592653589793]
```
Note that there is no rounding, so the *bare* numbers are used (as we do it with `Interval(0.1,0.1)`). 

Incidentally, there is no need to include something like `IReal`, which I proposed earlier.

lbenet:
Regarding merging this PR, I think we should push still a bit more to get more compatibility with the standard. I think what we have to do is have more control over the rounding of `Float64`s. I'm playing with this right now...

dpsanders:
I am not completely in agreement re converting immediately to the current precision, since that does not allow `Interval`s of e.g. `Rational`s to be manipulated exactly.

I don't see the necessity of changing the `Interval` constructor. Simply we write in the documentation, as is already there, that `@interval` is what you should use to create a correctly-rounded interval.

dpsanders:
What do you mean by better control over floating-point rounding? 

By the way, I nearly have the `crlibm` wrapper ready (in the branch `crlibm`), but there is something strange going on.

lbenet:
> I am not completely in agreement re converting immediately to the current precision, since that does not allow Intervals of e.g. Rationals to be manipulated exactly.

Since the changes involve only `Irrational` or `Integer`, rational intervals are constructed as they should be:
```julia
julia> Interval(1//2)
[1//2, 1//2]
```
Incidentally, currently `Interval(1//2)^2` throws an error...

dpsanders:
On current `master`, `Interval(1//2)^2` works fine, but `3*Interval(1//2)` gives an error. 

Maybe we should just give up on intervals containing non-floats?

dpsanders:
```julia
julia> promote(3, a)
ERROR: MethodError: `make_interval` has no method matching make_interval(::Type{Rational{Int64}}, ::Int64)`
```

lbenet:
> By the way, I nearly have the crlibm wrapper ready (in the branch crlibm), but there is something strange going on. 

I've noticed that you are pretty forward with that branch! What problem do you have there?

> What do you mean by better control over floating-point rounding? 

I'm referring to your [observation](https://github.com/dpsanders/ValidatedNumerics.jl/pull/55#issuecomment-139872508) about using `with_rounding` with `Float64`.

dpsanders:
Ah yes, I see. I really think those should go in a separate PR, since they represent fundamental changes to how the package operates.

dpsanders:
The problem in `crlibm` is some stupid metaprogramming stuff -- there is a loop over rounding modes to define the rounded functions in Julia, and only the last one in the list gets implemented; I have no idea why. You can try it out by following the instructions there.

lbenet:
> On current master, Interval(1//2)^2 works fine, but 3*Interval(1//2) gives an error.

You are right; the problem I get it on the current "ieee1788" branch... I'll check this now.

> Maybe we should just give up on intervals containing non-floats?

I agree on this, with regards to `Integer` and `Irrational` intervals; this is precisely what I was proposing. It would be nice to be able to keep rational intervals, but if this causes head-aches, it's not worth.

dpsanders:
OK, go ahead with getting rid of `Integer` and `Irrational` intervals. If we want to keep `Rational` ones, we should add some proper tests.

dpsanders:
On the other hand, the rational stuff can easily be done by any old interval cold written in a few lines with no rounding problems. So go ahead and eliminate that too...!

lbenet:
I'll leave interval rationals, until we really must get rid of them.

`Interval(1//2)^2` is working back again; stupid tiny thing in `mig` and `mag`... I'll add some tests, so this doesn't happen again...

lbenet:
I'm not sure why the tests fail in travis... In my machine, things seem to work correctly (v0.3 and 0.4).

dpsanders:
Is this ready to merge now?

lbenet:
Not yet. I'm retouching the `^` function, and the rounding macros. I'll see if I can get a better handling of the rounding for floats...

I know what you mean that the macros are quite hairy, literally...

dpsanders:
I think any changing of the macros really should go in  a separate PR. I
don't see any reason why everything should go in one monolithic PR...

lbenet:
You are right, it is probably a good idea to merge this, and continue in another PR with the control of rounding. So, if you feel this is ready, go ahead and merge it...

I just pushed a small commit, which adds some conversion rules to effectively restrict `Interval` to `Float64`, `BigFloat` or `Rational`, with tests for the constructors (in a separate file).

Just noticed that tests are not passing... Let me take a look first.

lbenet:
Something quite strange was happening with the tests. I included a new file for checking all constructors, which was being checked first, and then `consistency_tests.jl`. This had *a lot* of failed/erred tests. If a I change the order, so `consistency_tests.jl` is checked, and afterwards `construct_tests.jl`, the tests pass. Do you understand why? Could this be related with `set_bigfloat_precision(53)` in `consistency_tests.jl`? This shouldn't behave like this...

dpsanders:
Yes probably some interference from parameters that get set in previous tests. Try setting all the parameters as you would like in each file?

lbenet:
You're probably right; but that bothers me. Shall I do that in this PR, or later?

dpsanders:
As you prefer -  in this PR if it affects the tests.

lbenet:
Done!

lbenet:
I edited the title of this PR; WIP is erased

dpsanders:
So is it now ready to merge?

lbenet:
I think that the macros/constructors still need some revision, but this constitutes another PR.

Related to that, I think we need to increase the tests about the constructors, so the use of `Interval` yields a thin interval, while using `@interval` or the conversion methods should involve rounding (`:wide` or `:narrow`). For `:narrow` (and using `@interval` or the conversion methods) we should construct the thinnest possible interval.

The tests that still fail from the API of the [ITF1788 suite](https://github.com/oheim/ITF1788) are related to the implementation of the elementary functions. As we discussed somewhere above, this is at the end related to the rounding issues.

So, to avoid a monolitic PR, I suggest to merge this; I already have started something to improve the control of rounding.

dpsanders:
Done -- great work, thanks!

PR #56, opened 2015-09-20
Control rounding for many functions

lbenet:
This is some work to get tighter control of `Interval{Float64}`. It involves work in `make_interval`, in order to get as tight as possible intervals when `@interval(a)` is used. It also includes some handling of `@round`, and work to get the correct behavior of the arithmetic operations and the elementary function.


src/intervals/macro_definitions.jl
```diff
@@ -32,7 +32,3 @@ end
 macro biginterval(expr1, expr2...)
     make_interval(BigFloat, expr1, expr2)
 end
-
-macro make_interval(T, expr1, expr2...)
```
dpsanders: Why did you remove this macro? It allows to specify the type of the interval desired.

src/intervals/rationalize.jl
```diff
@@ -50,4 +50,5 @@ function old_rationalize{T<:Integer}(::Type{T}, x::AbstractFloat; tol::Real=eps(
     return convert(T,a)//convert(T,b)
 end
 
-old_rationalize(x::Union(Float64,Float32); tol::Real=eps(x)) = old_rationalize(Int, x, tol=tol)
+old_rationalize(x::Float64; tol::Real=eps(x)) = old_rationalize(Int, x, tol=tol)
+old_rationalize(x::Float32; tol::Real=eps(x)) = old_rationalize(Int, x, tol=tol)
```
dpsanders: These two definitions can be combined into one.

src/intervals/rounding.jl
```diff
@@ -12,6 +12,18 @@ if VERSION > v"0.4-"
 end
 
 
+macro show_rounding(T, expr)
+   quote
+       ee = @with_rounding($T, $expr, RoundNearest)
+       @show "NEA", ee
```
dpsanders: Better "Nearest"

src/intervals/rounding.jl
```diff
@@ -35,7 +47,26 @@ macro round(T, expr1, expr2)
...
         else # mode == :narrow
-            Interval(@with_rounding($T, $expr1, RoundDown), @with_rounding($T, $expr2, RoundUp))
+
+            # The idea here is, if RoundNearest is not equal to
+            # RoundDown or RoundUp, then the (Float64) operation does
+            # not round too sharply
+            lnea = @with_rounding($T, $expr1, RoundNearest)
+            ldow = @with_rounding($T, $expr1, RoundDown)
+            lup = @with_rounding($T, $expr1, RoundUp)
+            lo = ifelse(lnea != ldow && lnea != lup, lnea, ldow)
```
dpsanders: What is this line doing?

src/intervals/rounding.jl
```diff
@@ -44,17 +75,7 @@ end
 @doc doc"""`@thin_round` possibly saves one operation compared to `@round`.""" ->
```
dpsanders: This comment is no longer true after your change.

src/intervals/macro_definitions.jl
lbenet: Was never used...

dpsanders:
I can't check the `make_interval` stuff in detail right now. What's the overall plan?

src/intervals/macro_definitions.jl
dpsanders: True... But it seemed like it *might* be useful. Feel free to delete!

src/intervals/rationalize.jl
lbenet: I agree. However, `Union(..)` is deprecated in favor `Union{..}`, which has problems in version 0.3...

dpsanders: Yes, I think it will soon be time to stop supporting 0.3...

src/intervals/rounding.jl
lbenet: Ups... This was left to do some debugging. I'm not sure we want to keep it.

dpsanders: OK, delete it.

src/intervals/rounding.jl
lbenet: That's in the comment above :-) The idea is that the correct rounding should leave on of the roundings fixed to nearest, unless you really get the zero-distance for nearest rounding. It yields quite consistent things with respect to the IRF1788 tests.

lbenet:
The simple idea is to get the tightest possible interval for `@interval(a)` and the other related macros. 

The subtle thing, which we will need to discuss (but not now), is `make_interval(Float64,x::Irrational)`, since we use `BigFloats` to get the correct thing, and this is precision dependent. I decided to construct it with 53-bits `BigFloat`s, but that has consequences in `interval_parameters`, which I haven't been able to solve.

In my machine tests pass correctly... I'll check this tomorrow.

lbenet:
Problems with travis; see #59

lbenet:
In ma machine, with v0.4, tests pass...

lbenet:
New commit; now *all* tests in the ITF1788 suite pass, including all the currently defined trigonometric functions and (&^%$!@*) `^`. Travis is yielding some problems which I don't understand. Can you take a look on those problems to see what has to be done? 

Something to be done is certainly related to #59 ...

dpsanders:
Great work!!

This failure is https://github.com/JuliaLang/julia/issues/13399

I don't know what we can do about it.

dpsanders:
I have added functions of the form `sin(x, RoundUp)` etc. in the `mpfr_functions` branch of
[CRlibm.jl](https://github.com/dpsanders/CRlibm.jl) These should eventually go in Base but for now they are fine in `CRlibm.jl`. 

This will enable us to go back and simplify the code again. But this will need a new tagged version of CRlibm etc. For the moment we can go ahead with your current version.

lbenet:
It certainly would be nice if `sin(x, RoundDown)` works also for `BigFloat`s; for the time being I think it is fine to leave it like this.

lbenet:
I am implementing other functions to get as much compliance with the standard as we can...

dpsanders:
:+1:

dpsanders:
I think any actually *new* functions should go in a new PR.

lbenet:
Ok. So I'll stop it here; I already have the hyperbolic functions (`sinh` and `cosh`), since the remaining are not supported by CRlibm.

So this is ready for reviewing, but merging it should wait until we deprecate support of version 0.3.

dpsanders:
And put the Julia requirement in `REQUIRE`  as `julia 0.4-` for now.
(Or we can put 0.4 and wait til it comes out in a few days.)

dpsanders:
There are real test failures on Travis.
Do you have any idea what is happening?

Are these the same kind of inconsistent rounding errors that I was seeing a couple of months ago?

dpsanders:
The tests pass on my machine. These things are *really* annoying.

dpsanders:
I have rebased (without squashing, which I do not think is necessary) against current master (without any problems).

dpsanders:
I turned on Mac support on Travis and the tests are passing there!
So the problem is with Linux.

Do you have a Linux machine you can test this on?
I won't be able to do so until next week.

dpsanders:
https://travis-ci.org/dpsanders/ValidatedNumerics.jl/builds/84036509

dpsanders:
I wonder if this has to do with extended precision:

http://stackoverflow.com/questions/1961442/different-math-rounding-behaviour-between-linux-mac-os-x-and-windows

lbenet:
I do have access to a linux machine; I'll check the tests locally there, and I'll get back to you. It was precisely this kind of nonsense that was appearing at some stage, so at the end I was using `@pending` for those tests.

lbenet:
In the machine with linux that I have access, the tests pass (!?). This is the output from `versioninfo()`:
```julia
julia> versioninfo()
Julia Version 0.4.0-rc4+4
Commit 7048c40* (2015-10-07 00:33 UTC)
Platform Info:
  System: Linux (x86_64-suse-linux)
  CPU: AMD Opteron(tm) Processor 6376
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Piledriver)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.3
```

lbenet:
I get *identic* output (in Mac and Linux) for [your example](https://github.com/JuliaLang/julia/issues/13481#issuecomment-146283889) in https://github.com/JuliaLang/julia/issues/13481...

I really don't get why travis yields non-sense.

lbenet:
Tests are finally passing! Thanks @tkelman!

dpsanders:
Great! I haven't has a chance to review this in much detail, but if you think it's ready then go ahead and merge!

lbenet:
I think this is ready, though I prefer that you review it, since it includes changes in `make_interval`, and you have a really sharp eye :-)

dpsanders:
OK, give me a few days...

dpsanders:
Meanwhile you could add a few more tests for the new functionality, since coverage seems to have gone down.

lbenet:
Good point, I will include further tests.

lbenet:
Two tests are failing in travis (but not locally), again due to rounding issues, but now for Mac... I'll check this again later.

lbenet:
The tests that do not pass in travis on a mac are with `@pending`. It's not worth letting travis stop this any more.

lbenet:
Coveralls is still marginally reduced after the new tests...

lbenet:
Increased ~1% coverage, with some tests pending due to travis.

lbenet:
I just noticed an inconsistence in `inv` which is now corrected. (I also checked if the tests in the mac were not passing, but there are two still failing.)

REQUIRE
```diff
@@ -1,6 +1,6 @@
-julia 0.3
+julia 0.4-
```
dpsanders: Let's replace this with 0.4 now! :)

src/intervals/arithmetic.jl
```diff
@@ -85,10 +85,10 @@ function +{T<:Real}(a::Interval{T}, b::Interval{T})
 end
 +(a::Interval) = a
 
--{T<:Real}(a::Interval{T}) = @round(T, -a.hi, -a.lo)
+-(a::Interval) = Interval(-a.hi, -a.lo)
 function -{T<:Real}(a::Interval{T}, b::Interval{T})
     (isempty(a) || isempty(b)) && return emptyinterval(T)
-    a + (-b)  # @round(a.lo - b.hi, a.hi - b.lo)
```
dpsanders: I think `a + (-b)` will do the same thing, but no harm in being explicit.

src/intervals/arithmetic.jl
```diff
@@ -85,10 +85,10 @@ function +{T<:Real}(a::Interval{T}, b::Interval{T})
 end
 +(a::Interval) = a
 
--{T<:Real}(a::Interval{T}) = @round(T, -a.hi, -a.lo)
+-(a::Interval) = Interval(-a.hi, -a.lo)
```
dpsanders: Good catch that no rounding is required here.

src/intervals/arithmetic.jl
```diff
@@ -112,23 +112,19 @@ function *{T<:Real}(a::Interval{T}, b::Interval{T})
...
     isempty(a) && return emptyinterval(a)
 
-    T = eltype(a)
-    S = typeof(inv(a.lo))
     if in(zero(T), a)
-        a.lo < zero(T) == a.hi && return Interval{S}(-Inf, inv(a.lo))
-        a.lo == zero(T) < a.hi && return Interval{S}(inv(a.hi), Inf)
-        a.lo < zero(T) < a.hi && return entireinterval(S)
-        a == zero(a) && return emptyinterval(S)
+        a.lo < zero(T) == a.hi && return @round(T, -Inf, inv(a.lo))
```
dpsanders: Good catch.

src/intervals/arithmetic.jl
```diff
@@ -174,24 +170,62 @@ function /{T<:Real}(a::Interval{T}, b::Interval{T})
...
+    Interval(lo, hi)
+end
+
+
 ## Scalar functions on intervals (no directed rounding used)
 
 function mag(a::Interval)
     isempty(a) && return convert(eltype(a), NaN)
-    max( abs(a.lo), abs(a.hi) )
+    T = eltype(a)
```
dpsanders: Parametrize the function with `T` instead.

src/intervals/arithmetic.jl
```diff
@@ -174,24 +170,62 @@ function /{T<:Real}(a::Interval{T}, b::Interval{T})
...
+    r1, r2 = with_rounding(T, RoundUp) do
```
dpsanders: Not sure that rounding here is really necessary, since I guess `abs` just changes the sign, which should not involve rounding, and `max` also doesn't involve rounding. Still, I guess it doesn't hurt.

src/intervals/arithmetic.jl
```diff
@@ -174,24 +170,62 @@ function /{T<:Real}(a::Interval{T}, b::Interval{T})
...
+        abs(a.lo), abs(a.hi)
+    end
+    max( r1, r2 )
 end
+
 function mig(a::Interval)
     isempty(a) && return convert(eltype(a), NaN)
     zero(a.lo) ∈ a && return zero(a.lo)
-    min(abs(a.lo), abs(a.hi))
+    T = eltype(a)
```
dpsanders: Parametrize the function.

src/intervals/arithmetic.jl
```diff
@@ -227,17 +271,46 @@ union(a::Interval, b::Interval) = hull(a, b)
...
 
 function mid(a::Interval)
     isentire(a) && return zero(a.lo)
     (a.lo + a.hi) / 2
 end
+
 function diam(a::Interval)
     isempty(a) && return convert(eltype(a), NaN)
-    a.hi - a.lo
+    @with_rounding(eltype(a), a.hi - a.lo, RoundUp) #cf page 64 of IEEE1788
```
dpsanders: Parametrize the function.

src/intervals/conversion_promotion.jl
```diff
@@ -2,14 +2,12 @@
 
 ## Conversion and promotion
 
-## Default conversion is to Interval{Float64}
+## Default conversion to Interval, corresponds to Interval{Float64}
 convert{T<:Real}(::Type{Interval}, x::T) = make_interval(Float64, x)
```
dpsanders: Do we really need this "default" conversion?

src/intervals/conversion_promotion.jl
```diff
@@ -2,14 +2,12 @@
...
-@compat convert{T<:Union{Float64,BigFloat,Rational}}(::Type{Interval}, x::T) =
-    make_interval(T, x)
 
-## Conversion to intervals (with rounding) from Integer or Irrational
+## Conversion to specific type intervals
 @compat convert{T<:Union{Float64,BigFloat}, S<:Real}(::Type{Interval{T}}, x::S) =
```
dpsanders: `T <: AbstractFloat`

dpsanders: Despite Lint.jl, there's no need to parametrize `S` explicitly, you can just put `x::Real`, which is clearer in my opinion.

src/intervals/functions.jl
```diff
@@ -4,61 +4,140 @@
...
+    elseif a.hi ≤ zero(T)
+        !isthin(a) && return @controlled_round(T, a.hi^2, a.lo^2)
+        return a*a
+    else
+        return @controlled_round(T, mig(a)^2, mag(a)*mag(a))
+    end
+end
+
+# Floatingpoint power of a Float64/BigFloat interval:
+@compat function ^{T<:Union{Float64,BigFloat}}(a::Interval{T}, x::AbstractFloat)
```
dpsanders: `T <: AbstractFloat`

By the way, maybe we should check what happens if we try to make intervals of `Float32` or `Float16`.

src/intervals/functions.jl
```diff
@@ -4,61 +4,140 @@
 # Integer power:
 function ^{T<:Real}(a::Interval{T}, n::Integer)
```
dpsanders: I thought the idea with powers was to just convert `Float64` directly to `BigFloat` intervals with 53 bits of precision, do the power there, and then round back.

REQUIRE
lbenet: Done !

src/intervals/arithmetic.jl
lbenet: I must admit, that this one was actually caught by ITF1788...

src/intervals/arithmetic.jl
lbenet: Done; I've also implemented this to `mig`.

src/intervals/rounding.jl
```diff
@@ -35,26 +47,51 @@ macro round(T, expr1, expr2)
...
+            hi = @with_rounding($T, $expr2, RoundUp)
+            Interval(lo, hi)
         end
-
     end
 end
 
-@doc doc"""`@thin_round` possibly saves one operation compared to `@round`.""" ->
-macro thin_round(T, expr)
+macro controlled_round(T, expr1, expr2)
```
dpsanders: I'm afraid I still don't quite get the idea of this `controlled_round` macro.

src/intervals/rounding.jl
```diff
@@ -66,75 +103,99 @@ function split_interval_string(T, x::AbstractString)
...
-# but conversion from BigInt to BigFloat with correct rounding seems to be broken anyway # @thin_interval(BigFloat("$x"))
 
-make_interval(::Type{BigFloat}, x::Rational)  =  make_interval(BigFloat, x.num) / make_interval(BigFloat, x.den)
+# make_interval for BigFloat intervals
+make_interval(::Type{BigFloat}, x::AbstractString) =
+    split_interval_string(BigFloat, x)
+make_interval(::Type{BigFloat}, x::Irrational) = @thin_round(BigFloat, big(x))
+function make_interval(::Type{BigFloat}, x::Rational)
+    isinf(x) && return Interval(convert(BigFloat,x))
+    make_interval(BigFloat, x.num) / make_interval(BigFloat, x.den)
```
dpsanders: We can just do (I think)
```
make_interval(::Type{BigFloat}, x::Rational) = @thin_round(BigFloat, BigFloat(x))
```

src/intervals/conversion_promotion.jl
lbenet: I think this is the most reasonable thing, in the sense that usually we use `Float64`; in addition, this is along the same lines as `∅` defaults to `emptyinterval(Float64)`.

src/intervals/rounding.jl
```diff
@@ -66,75 +103,99 @@ function split_interval_string(T, x::AbstractString)
...
-make_interval(::Type{Float64}, x::Rational)  =  make_interval(Float64, x.num) / make_interval(Float64, x.den)
+# make_interval for Float64 intervals
+make_interval(::Type{Float64}, x::AbstractString) = split_interval_string(Float64, x)
+function make_interval(::Type{Float64}, x::Irrational)
+    a = with_bigfloat_precision(53) do
+        make_interval(BigFloat,x)
+    end
+    float(a)
+end
+function make_interval(::Type{Float64}, x::Rational)
```
dpsanders: Same comment as for `BigFloat`.

src/intervals/rounding.jl
```diff
@@ -66,75 +103,99 @@ function split_interval_string(T, x::AbstractString)
...
+    b = make_interval(Float64, x.hi)
+    Interval(a.lo, b.hi)
 end
 
-make_interval(::Type{Float64}, x::BigInt)    =  @thin_round(BigFloat, convert(Float64, x))
-make_interval(::Type{Float64}, x::BigFloat)  =  @thin_round(BigFloat, convert(Float64, x))
-make_interval(::Type{Float64}, x::Interval)  =  @round(BigFloat, convert(Float64, x.lo), convert(Float64, x.hi)) # NB: BigFloat to Float64 conversion uses *BigFloat* rounding mode
 
 
+# make_interval for Rational intervals
```
dpsanders: How are these used?

src/intervals/rounding.jl
```diff
@@ -187,7 +248,9 @@ function make_interval(T, expr1, expr2)
 end
 
 
-float(x::Interval) = make_interval(Float64, x)
+float(x::Interval) = Interval(convert(Float64,x.lo),convert(Float64,x.hi))
```
dpsanders: I don't think this is right. There should be some directed rounding possibly needed.

src/intervals/rounding.jl
```diff
@@ -187,7 +248,9 @@ function make_interval(T, expr1, expr2)
...
+# float(x::Interval) =
+#     @controlled_round(Float64, convert(Float64, x.lo), convert(Float64, x.hi))
```
dpsanders: Like this but just with `@round` I guess?

src/intervals/trigonometric_functions.jl
```diff
@@ -1,7 +1,9 @@
 # This file is part of the ValidatedNumerics.jl package; MIT licensed
 
-half_pi{T}(::Type{T}) = get_pi(T) / 2
-two_pi{T}(::Type{T})  = get_pi(T) * 2
+half_pi(::Type{Float64})  = @floatinterval(pi/2)
```
dpsanders: I'm not sure this is correct. `@floatinterval(pi) / 2` is safer.

src/intervals/trigonometric_functions.jl
```diff
@@ -1,7 +1,9 @@
...
+two_pi(::Type{Float64})   = @floatinterval(2pi)
+half_pi(::Type{BigFloat}) = @biginterval(pi/2)
+two_pi(::Type{BigFloat})  = @biginterval(2pi)
```
dpsanders: Similarly for all these.

dpsanders:
Great work, thanks!!

src/intervals/conversion_promotion.jl
dpsanders: But do we actually use this conversion anywhere? Can it just be removed?

dpsanders: I think the user has to be careful to specify to what kind of interval they want to convert. 
Alternatively, this information should be taken from the global `interval_parameters`

src/intervals/conversion_promotion.jl
lbenet: I have no problem with the observation about Lint.jl... 

Regarding `AbstractFloat`, do we really want to deal with (or care about) `Float32` or `Float16`? I didn't even think about this possibility before, but actually, `Interval(Float32(1.1))` yields `Interval{Float32}` as result, as it should. However, there is no `make_interval(Float32, x)`, and I am not sure that the rounding modes, or CRlibm care about this. I am in favor of concentrating only in `Interval{T}` with `T=Union{Float64, AbstractFloat, Rational}`, and I am always wondering about the very last one... 

We could certainly use `AbstractFloat` wherever is needed, but at some point the intervals will be promoted to Interval{Float64}, at least.

dpsanders: Yes, for the moment let's just worry about `Float32` and `Float64

src/intervals/functions.jl
lbenet: This is something I certainly tried, but it didn't quite worked; it was probably my implementation, but it simply was not yielding what it should (ITF1788).

dpsanders: OK, if it's passing the ITF1788 tests then that's a pretty good indication that it's correct!
Let's leave it for now.

By the way, I opened issue #65 about the need to incoporate the ITF1788 tests directly as part of our test suite somehow. This could be by making the ITF1788 suite a (test) dependency, for example. But that may be painful for Travis. Can we just copy over the files that ITF1788 outputs and add them to our test collection? (And then update when necessary.)

src/intervals/functions.jl
lbenet:
```julia
julia> Interval(Float32(1.1))
[1.1, 1.1]

julia> cos(ans)
[0.453596100177538, 0.45359610017753804]

julia> typeof(ans)
ValidatedNumerics.Interval{Float32}

julia> cos(Interval(Float32(1.1)))
[0.453596100177538, 0.45359610017753804]

julia> typeof(ans)
ValidatedNumerics.Interval{Float64}
```
So, `Interval{Float32}` yields something, which is a *literal* interval; `cos(...)` yields also something seemingly reasonable (but not quite!), but which is already an `Interval{Float64}`.

dpsanders:
I'm ready to merge this once you've fixed my minor comments!

src/intervals/arithmetic.jl
lbenet: I guess you are right; I will correct this and check it against ITF1788...

src/intervals/arithmetic.jl
lbenet: Done!

src/intervals/arithmetic.jl
lbenet: You are right; it was not necessary. Thanks!

src/intervals/functions.jl
lbenet: Some of the ITF1788 are really repetitive, but others are very sharp; maybe we could select few of them only...

src/intervals/rounding.jl
lbenet: The idea is to be sure that we round up and down only "one step"; their separation should be the local `eps`. The problem comes when the floating point is too close to the `RoundNearest` result, so they are *confussed* with that. In those cases, rounding up and down yields a difference which is two times the local `eps`. And you want to keep your results as tight as possible...

The above implementation is not the best regarding performance (six rounding operations and various changes in rounding mode); this could be improved (the trick we learned from Tucker), but once we are sure everything works correctly and as tight as possible!

src/intervals/rounding.jl
lbenet: I'll check it against the tests, and let you know in a moment...

lbenet: Yes, it works the way you propose, so I changed it to that. Thanks!

src/intervals/rounding.jl
lbenet: It seems to work as well....

lbenet: Yet, I just noticed that `@interval(1//0)` and `make_interval(Float64,1//0)` yield different numerical results, namely, `∅` and `[∞, ∞]`, respectively. The former is not correct (1//0 corresponds to `Inf` for `Rational`). Similar things occur with `@biginterval` and `make_interval(BigFloat, 1//0)`.

This is related to `transform`, which calls the actual operations among intervals, which is yields the `∅` if the interval in the denominator is the zero interval. I don't think this is too tragic...

src/intervals/rounding.jl
lbenet: Before, `transform` was handling everything, and if I recall correctly, a floating point was rationalized, and then the division of (integer) intervals was implemented. I simply maintained the possibility of working with `Interval{Rational}`. 
I think one example is the calculation of π (`calc_pi5`) in the tests.

src/intervals/rounding.jl
lbenet: If you round (further) then `float(@interval(0.1))` will not be equal to `@interval(0.1)`, as it is now, since it will be double rounded. This is tricky but important, since the trig functions (and #63) use it a lot! (Maybe, this was the reason I was not able to implement ^ using bigfloats with 53 bits, but I really forgot this.)

src/intervals/rounding.jl
lbenet: I think this also disturbs me, which may be the reason I left that thing commented there, but when I was using it, some kind of problem/inconsistency with the ITF1788 suite was appearing...

src/intervals/trigonometric_functions.jl
lbenet: I agree, it is *safer* (both contain pi/2), but for `Float64`, it is not *as tight as possible*:
```julia
julia> diam(@interval(pi)/2)
6.661338147750939e-16

julia> diam(@interval(pi/2))
2.220446049250313e-16
```
For bigfloat intervals, `@biginterval(pi)/2 == @biginterval(pi/2)` yields true, and corresponds to `@floatinterval(pi/2)`.

src/intervals/conversion_promotion.jl
lbenet: I think we simply define it, but don't use it, aside from the tests. Since the definition of these rules was already before, I guess I simply made them consistent with `make_interval`, after the changes in that function.

src/intervals/conversion_promotion.jl
lbenet: Maybe we should define `RestrictedFloat` to be `Union{Float64,BigFloat}`; that may make the code more readable.

As for `Rational`, the nice thing is precisely that there is no rounding!, so if you have to deal e.g. with polynomials with rational coefficients, the results do not need to be rounded! However, as soon as you include `sqrt` or things more involved, `Rational` intervals are not useful any more.

This is only to say that I have mixed feelings on this... I propose to keep it there for the time being.

lbenet:
Except for the part about `AbstractFloat` in this [comment](https://github.com/dpsanders/ValidatedNumerics.jl/pull/56#discussion-diff-41943442), your comments have been implemented in https://github.com/dpsanders/ValidatedNumerics.jl/commit/92ac0b314eb3e0b60dd5b0cbf3308b74c8088636. 

Tests pass (locally) as well as ITF1788 tests. Let's see what travis says...

lbenet:
And thanks a lot for the very careful review!

PR #60, opened 2015-10-06
Update NEWS.md

lbenet:
Adds some description in `NEWS.jl` of changes related to #55; it also includes a notice that 0.1.3 will be the last version which will be compatible with Julia 0.3

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # What's new in ValidatedNumerics.jl
 
+# 0.2
+
+# 0.2.0
+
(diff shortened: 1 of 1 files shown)


dpsanders:
Please move the NEWS.md change with respect to 0.2 to PR #56.

lbenet:
I have erased mentioning 0.2 here, and I will mention it in #56. But I prefer to wait a bit (until this is merged) to do an easy rebase (instead of doing something more involved). This might be a good chance to squash some commits of #56.

dpsanders:
Thanks. I've tagged 0.1.3.

Issue #67, opened 2015-10-14
IEEE 1788 conformance


dpsanders:
It would be very useful to have a check list of functions that still need to be implemented for conformance with the IEEE 1788 standard for interval arithmetic.

Cc @lbenet

lbenet:
The following is a list of *required* functions to be IEEE 1788 conforming, following [Nehmeier's paper](http://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=6893854) (cf. sect. II.C.1, "Required operations for all flavours"):

- [x] Basic operations: `neg` (`-`), `add` (`+`), `sub` (`-`), `mul` (`*`), `div` (`/`), `recip` (`inv` in Julia), `sqr`, `sqrt`, `fma`; [#56](https://github.com/dpsanders/ValidatedNumerics.jl/issues/56)

- [ ] Case function: `case`

- [ ] Power functions: `pown` (`^`), `pow` (`^`), `exp`, `exp2`, `exp10`, `log`, `log2`, `log10`. Partially implemented in [#56](https://github.com/dpsanders/ValidatedNumerics.jl/issues/56). 
**Needed**: `exp2`, `exp10`, `log2`, `log10`. (Maybe some metaprogramming is useful here.)

- [ ] Trigonometric/hyperbolic functions: `sin`, `cos`, `tan`, `asin`, `acos`, `atan`, `atan2`, `sinh`, `cosh`, `tanh`, `asinh`, `acosh`, `atanh`. Almost totally implemented in [#56](https://github.com/dpsanders/ValidatedNumerics.jl/issues/56) and [#63](https://github.com/dpsanders/ValidatedNumerics.jl/issues/63).
**Needed**:  `atan2`

- [ ] Integer functions: `sign`, `ceil`, `floor`, `trunc`, `roundTiesToEven`, `roundTiesToAway`. Almost totally implemented in [#56](https://github.com/dpsanders/ValidatedNumerics.jl/issues/56).
**Needed**: `roundTiesToEven` and `roundTiesToAway` remain to be implemented.

- [x] Absmax functions: `abs`, `min`, `max`; [#56](https://github.com/dpsanders/ValidatedNumerics.jl/issues/56) 

- [ ] Cancel functions: `cancelPlus` and `cancelMinus`

- [x] Set operations: `intersection` (`intersect`, `∩`), `convexHull` (`hull`)

[...]

PR #68, opened 2015-10-15
Add functions to get compatibility with IEEE 1788

lbenet:
Add atan2, and use Interval(big(...),big(...)) in ^

The second one yields the same results without the penalty of rounding.


lbenet:
I've pushed some new commits with tests, which are related with what is required for conformance with the IEEE1788. I have also updated the list in #67 to refer to the concrete commit where the implementation was made. Including the last commit, I think we have all functions we needed.

@dpsanders Review carefully `cancelminus` (`cancelplus` is the same, slightly disguised), since some evaluation in extended precision is required, even for `Interval{BigFloat}`. The cases of interest are of the form `cancelminus( Interval(-1.0, a), Interval(-b, 1.0) )`, where a, b are *positive* floating points, less  than `eps(1.0)`.

lbenet:
Mmmm... I forgot actually to include tests for `Interval{BigFloat}` for the case which requires particular attention.

lbenet:
Pushed new tests to increase coverage, including checking two tests which were `@pending` (due to Mac issues). 

If the tests do not pass, maybe we should turn off mac checking in travis; I guess this is due to the docker version we are using (because the problem was not there before this change). Needless to say, but the tests pass locally.

lbenet:
I turned off the support for Mac; tests pass, and coverage has been increased to ~95%

src/intervals/functions.jl
```diff
@@ -84,7 +84,7 @@ end
         hi = hull(hi, hi1)
     else
         lo, hi = with_bigfloat_precision(53) do
-            aa = @biginterval(a)
+            aa = Interval(big(a.lo), big(a.hi)) # before: @biginterval(a)
```
dpsanders: If `a` is a `BigFloat` interval, we don't want to restrict its precision to 53 bits.

src/intervals/functions.jl
```diff
@@ -159,26 +159,44 @@ function exp(a::Interval{Float64})
...
 end
 
-function exp(a::Interval{BigFloat})
-    isempty(a) && return a
-    @controlled_round(BigFloat, exp(a.lo), exp(a.hi))
+for fn in (:exp2, :exp10)
+    @eval begin
+        function ($fn)(a::Interval{Float64})
+            isempty(a) && return a
+            res = with_bigfloat_precision(53) do
```
dpsanders: I think we should introduce a `big53` function.
Then you can write something like the following to define this kind of function.

```julia
exp(a::Interval{Float64}) = float(exp(big53(a)))
```

src/intervals/trigonometric_functions.jl
```diff
@@ -4,9 +4,13 @@ half_pi(::Type{Float64})  = @floatinterval(pi/2)
 two_pi(::Type{Float64})   = @floatinterval(2pi)
 half_pi(::Type{BigFloat}) = @biginterval(pi/2)
 two_pi(::Type{BigFloat})  = @biginterval(2pi)
-
 half_pi{T<:AbstractFloat}(x::T) = half_pi(T)
 
+pi_interval{T<:Real}(::Type{T}) = get_pi(Float64)
```
dpsanders: This definition is incorrect for `BigFloat`.

src/intervals/trigonometric_functions.jl
```diff
@@ -4,9 +4,13 @@ half_pi(::Type{Float64})  = @floatinterval(pi/2)
...
+range_atan2{T<:Real}(::Type{T}) = Interval(-pi_interval(T).hi, pi_interval(T).hi)
+half_range_atan2(::Type{Float64}) = @floatinterval(-pi/2, pi/2)
+half_range_atan2(::Type{BigFloat}) = @biginterval(-pi/2, pi/2)
```
dpsanders: This is inconsistent with the above definition of `range_atan2` for `BigFloat`.

dpsanders: This seems to be the correct way of defining the range, in the sense that it correctly changes if the interval precision changes.

dpsanders:
This looks good, thanks! Could you please fix the comments about the definitions of trig ranges and then I think we can merge this.

src/intervals/arithmetic.jl
```diff
@@ -318,6 +346,38 @@ function radius(a::Interval)
     max(m-a.lo, a.hi-m)
 end
 
+# cancelplus and cancelminus
+@doc doc"""
+`cancelminus(a, b)` returns the unique interval `c` such that `b+c=a`.""" ->
+function cancelminus(a::Interval, b::Interval)
+    T = promote_type(eltype(a), eltype(b))
```
dpsanders: I think you should be able to do this by just defining
```julia
function cancelminus{T}(a::Interval{T}, b::Interval{T})
```
and relying on the promotion rules we already have defined. Similarly for `fma` etc.

src/intervals/trigonometric_functions.jl
dpsanders: We should have either `get_pi` or `pi_interval`. This function should return the current value of `interval_parameters.pi`, and all of the other `pi` functions should be defined in terms of this.

dpsanders:
Merging so that we can carry on working on top of this.

PR #77, opened 2015-11-01
Clean up functions to use CRlibm syntax

dpsanders:
Rewrite to use functions of the form `sin(x::BigFloat, RoundDown)` when possible.
Rewrite `^` to convert `Float64` intervals straight into `BigFloat` intervals.

Introduces `big53` and `to_float` functions to convert `Float64` to `BigFloat` intervals and back.

Also some other clean-up and docs.

`@controlled_round` is now no longer used.

### DIFF
--- a/src/intervals/conversion_promotion.jl
+++ b/src/intervals/conversion_promotion.jl
@@ -3,18 +3,19 @@
 ## Conversion and promotion
 
 ## Default conversion to Interval, corresponds to Interval{Float64}
+# do we really need this?
 convert{T<:Real}(::Type{Interval}, x::T) = make_interval(Float64, x)
 
 ## Conversion to specific type intervals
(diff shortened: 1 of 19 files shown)


src/intervals/conversion_promotion.jl
```diff
@@ -3,18 +3,19 @@
 ## Conversion and promotion
 
 ## Default conversion to Interval, corresponds to Interval{Float64}
+# do we really need this?
```
lbenet: We *do* need this! For instance, when you calculate `x+0.1` with `x` an Interval, `0.1` is promoted from float to interval, which at some point uses `convert{Interval{Float64},0.1}` an thus yields `Interval[0.09999999999999999, 0.1]` (and not `Interval(0.1)`).

dpsanders:
Superseded by #78

PR #79, opened 2015-11-03
Add alias mechanism for IEEE 1788 function names.

dpsanders:
Usage: @alias convexHull


dpsanders:
This turned out to be harder than expected.
The easiest thing to do would be to just do something like

```julia
const convexHull = hull
```

The point of defining the macro is that if we decide to go down this route, it will not require much more work.

dpsanders:
Since there did not seem to be any enthusiasm for this idea, I'm going to close it.

PR #86, opened 2016-02-10
Remove empty rounding operations for non-floats. Clarify in NEWS that non-floats are no longer allowed.

dpsanders:


dpsanders:
Closed in favor of #87

PR #87, opened 2016-02-11
Fix rounding using rational intervals

dpsanders:
e.g. sqrt(Interval(1//3))


dpsanders:
`sqrt(Interval(1//3))` now rounds correctly.

dpsanders:
This is fine on 0.4 and needed for the changes to make 0.5 work, so merging...

PR #88, opened 2016-02-11
Basic support for multidimensional IntervalBox's

dpsanders:
[`FixedSizeArrays.jl`](https://github.com/SimonDanisch/FixedSizeArrays.jl) makes this really easy.


dpsanders:
@lbenet I would like to merge this and tag a new version. Travis on 0.5 is not working for the moment. Is this OK with you?

lbenet:
Sorry, I've been too busy lately. Give me today to take a quick look on this, and if i don't react, just go ahead and merge.

REQUIRE
```diff
@@ -1,5 +1,4 @@
 julia 0.4
-FactCheck 0.3
```
lbenet: Don't we need this here, to point out that the package uses this? I see that it appears in `test/REQUIRE`, but isn't it necessary here to install it as a dependency when using `Pkg.add("ValidatedNumerics")`?

src/ValidatedNumerics.jl
```diff
@@ -62,8 +70,13 @@ end
 
 include("misc.jl")
 include("intervals/intervals.jl")
+
+a..b = @interval(a, b)
+
+include("multidim/multidim.jl")
 include("root_finding/root_finding.jl")
 
+Base.isfinite(x::Interval) = isfinite(x.lo) && isfinite(x.hi)
```
lbenet: I think this is not necessary here, since you are including this function in `src/intervals/special_intervals.jl`.

src/intervals/special_intervals.jl
```diff
@@ -33,6 +33,7 @@ nai() = nai(get_interval_precision()[1])
 
 isnai(x::Interval) = isnan(x.lo) || isnan(x.hi)
 
+isfinite(x::Interval) = isfinite(x.lo) && isfinite(x.hi)
```
lbenet: Since you are including `isfinite` here, I think it would be worth to also include `isnan()` and `isinf()`, which are somehow missing in ValidatedNumerics (and are necessary to a nicer experience when combining this with TaylorSeries.jl). We already have something that implements this functionality, but these functions are more julian.
As far as I can tell, `isnan()` is equivalent to `isnai()`, and `isinf()` is equivalent to `isunbounded()` (or to the negation of `isfinite()`). Could you include them here?

src/multidim/box_macro.jl
```diff
@@ -0,0 +1,42 @@
...
+
+    @box f(x, y) = (x + y, x - y)
+
+    X = IntervalBox(1..1, 2..2)
+    f(X)
+
+(No significant error checking is performed!)
+"""
+
+macro box(ex)
```
lbenet: Could we rename @box` to `@intervalbox` or `@boxinterval`, or whatever name that points out that we are creating an *interval* box? At the moment we have `@biginterval`, which I like because it's pretty clear what it does.

lbenet:
LGTM. Aside from the comments, which are minor things, I think we should wait to tag the new version to include some of the things related to #91 (though perhaps not as in #92).

REQUIRE
dpsanders: No, it will be installed automatically when `Pkg.test` is run, and then deleted again. This is the recommended method for test-only dependencies.

lbenet: Thanks for commenting on this. I didn't know!

src/ValidatedNumerics.jl
dpsanders: Thanks, fixed.

src/intervals/special_intervals.jl
dpsanders: `isinf` and `isnan` already work by Julia magic!

src/multidim/box_macro.jl
dpsanders: I wanted to use the name `Box`, instead of `IntervalBox`, but until recently `Box` was an exported internal name. However, maybe `IntervalBox` is a better name. I will change `@box` to `@intervalbox`

src/intervals/special_intervals.jl
dpsanders: `isnan` is now the same as `isnai`. In any case, this seems to be a pretty useless thing to do, since I don't see how `Interval`s of `NaN`s will get created ever.

dpsanders:
I think this is ready.

src/intervals/special_intervals.jl
lbenet: > `isinf` and `isnan` already work by Julia magic!

Last time I checked (playing together with TaylorSeries.jl, in julia 0.4) it was not working... 

> `isnan` is now the same as `isnai`. In any case, this seems to be a pretty useless thing to do, since I don't see how `Interval`s of `NaN`s will get created ever.

Probably you are right. Yet, it is part of IEEE1768... The name `isnai`, if I recall correctly, was taken from the standard.

lbenet:
There are failing tests (julia 0.4); it actually didn't get to the ITF1788 test suite. 

I'm checking this locally.

lbenet:
They are also failing locally (for both 0.4.4 and 0.5-dev), precisely on "Operations with boxes".

src/multidim/box.jl
```diff
@@ -0,0 +1,34 @@
+# Multidimensional intervals, called IntervalBox
+
+#using ValidatedNumerics
+#using FixedSizeArrays
+
+import Base:
+    ⊆, ∩, isempty
+
+export IntervalBox
```
lbenet: `IntervalBox` is exported in src/ValidatedNumerics.jl; I don't think it needs to be exported here.

test/multidim_tests/multidim_tests.jl
```diff
@@ -0,0 +1,31 @@
...
+
+    v = [@interval(i, i+1) for i in 1:10]
+    V = IntervalBox(v...)
+
+    @fact length(V) --> 10
+
+end
+
+facts("@box tests") do
+    @box f(x, y) = (x + y, x - y)
```
lbenet: I think this should be `@intervalbox`...

lbenet: I think this is the problem with the tests. Can you correct this?

src/multidim/box.jl
dpsanders: Thanks

dpsanders:
Sorry, should be OK now.

lbenet:
Tests for Julia 0.4 are passing; so I will merge (in some 15 minutes). I want to see if the tests in 0.5 will finish or not.

lbenet:
I noticed that this cannot by merged due to a conflict. Can you solve it? 
(Tests in travis 0.5 are not passing, with the same all problem...)

dpsanders:
Finally ready.

dpsanders:
Maybe I should squash?

lbenet:
Let's leave this as it is. I'm merging!

lbenet:
It's really a nice PR

PR #89, opened 2016-02-12
Update names of precision and rounding functions.

dpsanders:
Uses the new names from Julia v0.5. Needs https://github.com/JuliaLang/Compat.jl/pull/173


dpsanders:
Tests now pass (finally) on both 0.4 and 0.5.
There is an annoying bit of version testing in one file (`rounding.jl`); I couldn't work out how to get round this using `Compat.jl`. Merging...

Issue #91, opened 2016-03-10
Type unstability in `make_interval`


lbenet:
I just became across an issue that suggests that there is somewhere/somehow type instability in ValidatedNumerics.jl. The following example illustrates the problem:

```julia
julia> f(x) = 1+x
f (generic function with 1 method)

julia> @code_warntype f(Interval(1.0))
Variables:
  x::ValidatedNumerics.Interval{Float64}

Body:
  begin  # none, line 1:
      GenSym(2) = (ValidatedNumerics.make_interval)(Float64,1)::Any
      GenSym(1) = call(ValidatedNumerics.Float64,(top(getfield))(x::ValidatedNumerics.Interval{Float64},:lo)::Float64,ValidatedNumerics.RoundDown)::Float64
      GenSym(0) = call(ValidatedNumerics.Float64,(top(getfield))(x::ValidatedNumerics.Interval{Float64},:hi)::Float64,ValidatedNumerics.RoundUp)::Float64
      return GenSym(2) + call(ValidatedNumerics.Interval{Float64},GenSym(1),GenSym(0))::ValidatedNumerics.Interval{Float64}::Any
  end::Any
```

Note the `Any`s appearing in `GenSym(2)`, that suggest looking at `make_interval()`, and going further, I think that `split_interval_string()` may be responsible.

Using, `f(x)=x`, or `exp(x)`, is type stable.

dpsanders:
But it finds the correct type in the end for `x`, so I don't think there's a real type instability.
Apparently it's known that this can happen.

lbenet:
I'm not sure I understand what you mean; according to the output, f(x) is `Any`, instead of being `Interval`, which is what matters for the compiler. I think the desired outcome mimics what is obtained for x being a `Float64`, which is a `Float64`. For the application where I found this behavior, it does induce a performance penalty.

I'll try to push later what I've done so far; at least for this case, it solves the problem.

dpsanders:
I can see that `split_interval_string` could be type unstable, you're right.
But it seems to get the type right for `f`:

```julia
julia> f(Interval(1.0))
[2.0, 2.0]

julia> typeof(ans)
ValidatedNumerics.Interval{Float64}
```
Or what am I missing?

lbenet:
Ok, but this is not what the compiler gets. (The issue is like the problem to pass functions as arguments of functions: as I understand it, your function may be ok, but the compiler doesn't know it.)

What we should aim is to get something like
[...]

PR #92, opened 2016-03-11
Solves some type instability issues, or diminishes its effects

lbenet:
Aims to solve, or at least diminish the consequences noticed in #91

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -191,20 +191,20 @@ function fma{T}(a::Interval{T}, b::Interval{T}, c::Interval{T})
         return entireinterval(T)
     end
 
-    lo = setrounding(T, RoundDown) do
-        lo1 = fma(a.lo, b.lo, c.lo)
-        lo2 = fma(a.lo, b.hi, c.lo)
-        lo3 = fma(a.hi, b.lo, c.lo)
-        lo4 = fma(a.hi, b.hi, c.lo)
-        min(lo1, lo2, lo3, lo4)
-    end :: T
-    hi = setrounding(T, RoundUp) do
-        hi1 = fma(a.lo, b.lo, c.hi)
-        hi2 = fma(a.lo, b.hi, c.hi)
-        hi3 = fma(a.hi, b.lo, c.hi)
-        hi4 = fma(a.hi, b.hi, c.hi)
-        max(hi1, hi2, hi3, hi4)
-    end :: T
+    setrounding(T, RoundDown)
+    lo1 = fma(a.lo, b.lo, c.lo)
+    lo2 = fma(a.lo, b.hi, c.lo)
+    lo3 = fma(a.hi, b.lo, c.lo)
+    lo4 = fma(a.hi, b.hi, c.lo)
+    lo = min(lo1, lo2, lo3, lo4)
+    setrounding(T, RoundUp)
+    hi1 = fma(a.lo, b.lo, c.hi)
+    hi2 = fma(a.lo, b.hi, c.hi)
+    hi3 = fma(a.hi, b.lo, c.hi)
+    hi4 = fma(a.hi, b.hi, c.hi)
+    hi = max(hi1, hi2, hi3, hi4)
+    setrounding(T, RoundNearest)
+    
     Interval(lo, hi)
 end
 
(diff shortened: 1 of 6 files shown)


lbenet:
I tried to solve this by annotating the expected type in different places.

lbenet:
I don't like the solution (type annotations) either, but at least solves the issue partially. I agree that not being too careful about what we construct, leads to this problem. I simply provided this solution, because it makes my computations nearly a factor 2 faster. (I still need to make other improvements.)

lbenet:
Tests actually work fine in Julia v0.5, but take 10x more!

lbenet:
Part of the problem of loosing type stability is related to the fact that some macros/functions, in a way, pass a function as an argument. This happens often when we are changing the rounding mode. 

I just pushed a new commit (that only deals with the problematic functions in `intervals/arithmetic.jl`, i.e., `fma`, `mig` and `diam`) which yields type-stable code, which is actually faster than what I did previously.

dpsanders:
Passing functions as arguments is no longer slow in 0.5 since https://github.com/JuliaLang/julia/pull/13412, so I'm not sure those changes are necessary.

lbenet:
I've done some tests (only for `fma()`, `mig()` and `diam()`, which were corrected in the last commit), using julia 0.4 and 0.5, and considering the current master (commit bf62bf4) and b1f96ba, using `Interval{Float64}` and `Interval{BigFloat}`. The results below are for 10^5 iterations of those functions (in that order).

- JULIA 0.4
In master:
Testing: Float64
  2.459652 seconds (53.00 M allocations: 1007.090 MB, 4.59% gc time)
  0.586216 seconds (13.00 M allocations: 289.917 MB, 3.65% gc time)
  0.104019 seconds (6.00 M allocations: 137.329 MB, 10.64% gc time)
Testing: BigFloat
  8.345536 seconds (56.00 M allocations: 1.669 GB, 17.86% gc time)
  2.010344 seconds (18.00 M allocations: 564.575 MB, 22.14% gc time)
  0.483570 seconds (6.00 M allocations: 175.476 MB, 24.32% gc time)
For b1f96ba:
Testing: Float64
  0.033382 seconds (149 allocations: 10.167 KB)
  0.020198 seconds (5 allocations: 176 bytes)
  0.014262 seconds (5 allocations: 176 bytes)
Testing: BigFloat
  5.557482 seconds (48.00 M allocations: 1.460 GB, 24.38% gc time)
  1.179876 seconds (13.00 M allocations: 411.987 MB, 33.96% gc time)
  0.399777 seconds (4.00 M allocations: 114.441 MB, 28.36% gc time)

- JULIA 0.5
In master:
Testing: Float64
  0.198245 seconds (2.00 M allocations: 30.528 MB, 19.88% gc time)
  0.081418 seconds (1.00 M allocations: 30.518 MB, 3.26% gc time)
  0.062971 seconds (1.00 M allocations: 15.259 MB, 3.84% gc time)
Testing: BigFloat
  5.417516 seconds (36.00 M allocations: 1.311 GB, 23.16% gc time)
  1.070972 seconds (11.00 M allocations: 396.729 MB, 34.54% gc time)
  0.431670 seconds (4.00 M allocations: 114.441 MB, 24.42% gc time)
For b1f96ba:
Testing: Float64
  0.040103 seconds (160 allocations: 11.152 KB)
  0.019922 seconds (5 allocations: 176 bytes)
  0.014244 seconds (5 allocations: 176 bytes)
Testing: BigFloat
  5.095795 seconds (34.00 M allocations: 1.252 GB, 23.44% gc time)
  0.963879 seconds (9.00 M allocations: 350.952 MB, 35.31% gc time)
  0.340945 seconds (3.00 M allocations: 99.182 MB, 26.63% gc time)

Comparing Julia 0.4 vs 0.5, 0.5 is better (both in master and in b1f96b). 

Comparing master vs b1f96b, I think b1f96b does something *really* good, both in elapsed time as in the allocated memory, and for both Interval cases considered, in particular for `Float64`s.

lbenet:
So I agree with your comment above, that 0.5 solves many things, but b1f96ba too.

dpsanders:
I think the origin of the problem is in this kind of thing:

```
julia> x = @interval(1)
[1.0, 1.0]

julia> @code_warntype convert(Interval{BigFloat}, x)
Variables:
  #s15::Type{ValidatedNumerics.Interval{BigFloat}}
  x::ValidatedNumerics.Interval{Float64}

Body:
  begin  # /Users/dpsanders/.julia/v0.4/ValidatedNumerics/src/intervals/conversion_promotion.jl, line 10:
      return (ValidatedNumerics.make_interval)(T,x::ValidatedNumerics.Interval{Float64})::Any
  end::Any
```

dpsanders:
I don't see how to fix it though. However, if in 0.5 this is no longer a problem, then I don't think it's worth worrying too much about?

lbenet:
I agree that somewhere in `src/rounding.jl` there is something to be done (`convert` eventually calls `make_interval`). In your example above, it calls `@round`, which then calls `@setrounding`. The latter macro essentially passes a function as an argument to a function. That is the reason that i annotated `@setrounding` with the expected type.

Though I think this is important (see the benchmarks), if you prefer we can leave this for now, and I'll take a look on this. So we can tag a new version as you wanted.

dpsanders:
Since the type instability no longer occurs on 0.5 after the functions rewrite, I am going to close this. If you don't agree, feel free to reopen.

PR #98, opened 2016-03-26
Change make_interval to convert

dpsanders:
Replace `make_interval` by `convert`. Includes significant simplification of `convert` but also changes how some intervals are constructed, which led to some rejigging of tests.

Also showed up a bug with `with_interval_precision` that did not correctly change the precision back afterwards.

Split up some of the longer test suites.


src/intervals/conversion_promotion.jl
```diff
@@ -19,3 +11,103 @@ promote_rule{T<:Real, S<:Real}(::Type{Interval{T}}, ::Type{S}) =
 
 promote_rule{T<:Real}(::Type{BigFloat}, ::Type{Interval{T}}) =
     Interval{promote_type(T, BigFloat)}
+
+
+## Conversion rules
+
+# Just specifying Interval defaults to Float64 interval
+# Remove this default so that it is necessary to explicitly specify?
+convert(::Type{Interval}, x::Real) = convert(Interval{Float64}, x)
```
lbenet: I think a reasonable default is: 

- for `x::Rational{T}` create an `Interval{Rational{T}}`;
- for `x::BigFloat`, create `Interval{BigFloat}`; 
- for the rest of `AbstractFloat`s create `Interval{Float64}`.

If someone wants to do something else more specific, it should be explicitly given.

lbenet:
Some of the `convert` methods that now are included in `src/conversion_promotion.jl` (which is the place where they should be!) are still in `src/rounding.jl`. This also includes the function `split_interval_string`. Some cleaning dups is needed.

src/intervals/rounding.jl
```diff
@@ -81,81 +85,80 @@ end
...
+convert{T<:Integer, S<:BigFloat}(::Type{Interval{Rational{T}}}, x::S) =
     Interval(rationalize(T, x))
-function make_interval{T<:Integer}(::Type{Rational{T}}, x::Interval)
-    a = make_interval(Rational{T}, x.lo)
-    b = make_interval(Rational{T}, x.hi)
-    Interval(a.lo, b.hi)
-end
 
 
 doc"""`transform` transforms a string by applying the function `f` and type
```
lbenet: Probably a better place to have `make_interval` and `transform` is in `src/macros`, since they are linked to `@interval` and its relatives.

test/interval_tests/construct_tests.jl
```diff
@@ -52,36 +49,43 @@ facts("Constructing intervals") do
...
     # Constructors from the macros @interval, @floatinterval @biginterval
     set_interval_precision(53)
+
     a = @interval(0.1)
     b = @interval(pi)
+
     @fact nextfloat(a.lo) --> a.hi
     @fact typeof(a) --> Interval{BigFloat}
     @fact a --> @biginterval("0.1")
-    @fact float(a) --> @floatinterval(0.1)
```
lbenet: Since `float{Interval}` exists, we may keep this test as well, even if it does the same as the next line.

test/interval_tests/numeric_tests.jl
```diff
@@ -73,9 +73,9 @@ facts("Numeric tests") do
     @fact @biginterval(-3,4) ^ 0.5 --> @biginterval(0, 2)
 
     @fact @interval(1,27)^@interval(1/3) --> roughly(Interval(1., 3.))
-    @fact @interval(1,27)^(1/3) --> Interval(1., 3.)
-    @fact @interval(1,27)^(1//3) --> Interval(1., 3.)
-    @fact @interval(0.1,0.7)^(1//3) --> Interval(0.46415888336127786, 0.8879040017426009)
+    @fact @interval(1,27)^(1/3) --> roughly(Interval(1., 3.))
```
lbenet: Why is this test change needed? Before we had the nice and correct behavior that `@fact @interval(1,27)^(1/3) --> Interval(1., 3.)` was `true`.

test/interval_tests/numeric_tests.jl
```diff
@@ -73,9 +73,9 @@ facts("Numeric tests") do
...
+    @fact Interval(1., 3.) ⊆ @interval(1,27)^(1//3) --> true
+    @fact @interval(0.1,0.7)^(1//3) --> Interval(0.4641588833612778, 0.8879040017426008)
```
lbenet: I dislike (and worries me) the necessity to adapt the *numeric* values of the tests (e.g., line 78, and others below). While the present may well be correct, some examples should that the *new* intervals are slightly larger (or shifted!) in comparison to what we had before. 

So, are we correctly rounding, and using tests that yield the tightest interval?

dpsanders:
Note that at least one of the errors with the current version is due to [this bug in Julia](https://github.com/JuliaLang/julia/issues/15659).

dpsanders:
Some power test results have now changed again due to the new (simpler and more coherent) definitions of `convert` (the old `make_interval`). 

In any case it seems to me that there are some problems with the power code that I will look at later.

src/intervals/conversion_promotion.jl
dpsanders: What you are suggesting for rationals and `BigFloat` is already done by the constructor:

```
julia> x = 1//3
1//3

julia> Interval(x)
[1//3, 1//3]
```

You only need to explicitly call `convert` to convert it to some other type of interval:
```
julia> convert(Interval{Float64}, x)
[0.3333333333333333, 0.33333333333333337]
```
In that case, you would want to be explicit about the type you were converting to, I feel.

src/intervals/rounding.jl
dpsanders: Yes, I agree. I'm planning to do some refactoring like that.

dpsanders:
@lbenet Tests finally pass on 0.4. This turned out to be more work than I was expecting! I think the result is nice. I suggest we merge this as it currently is and then do more refactoring later.

src/intervals/conversion_promotion.jl
lbenet: Using your example, for `x=1//3` my point is that `convert(Interval,x)` should default in `Interval{typeof(x)}(x)`, but it doesn't:
```julia
julia> x = 1//3
1//3

julia> convert(Interval, x)
[0.3333333333333333, 0.33333333333333337]
```

src/intervals/hyperbolic_functions.jl
```diff
@@ -56,6 +56,6 @@ for f in (:tanh, :asinh, :acosh, :atanh)
     @eval function ($f)(a::Interval{Float64})
         isempty(a) && return a
 
-        float(($f)(big53(a)))
+        float( $f (big53(a)) )
```
lbenet: I get the following warning due to this line:
```
WARNING: deprecated syntax "$f (" at /Users/benet/Fisica/6-IntervalArithmetics/ValidatedNumerics.jl/src/intervals/hyperbolic_functions.jl:59.
Use "$f(" instead.
```
I guess the blank space is bothering...

src/intervals/rounding.jl
```diff
@@ -238,3 +152,5 @@ function set_interval_rounding(mode)
 
     parameters.rounding = mode  # a symbol
 end
+
+big{T}(x::Interval{T}) = convert(Interval{BigFloat}, x)
```
lbenet: Shouldn't `big` be imported from base (`Base.big`)? Otherwise, we do not add a new method...

test/interval_tests/trig_tests.jl
```diff
@@ -75,7 +74,7 @@ facts("Trig tests") do
 
     @fact atan2(∅, entireinterval()) --> ∅
     @fact atan2(entireinterval(), ∅) --> ∅
-    @fact atan2(@interval(0.0, 1.0), @biginterval(0.0)) --> @biginterval(pi/2)
+    @fact atan2(@interval(0.0, 1.0), @biginterval(0.0)) --> @interval(pi/2)
```
lbenet: Why is this change in the tests required? I guess this is related to the fact that
```julia
julia> atan2(@interval(0.0, 1.0), @biginterval(0.0))
[1.5707963267948966, 1.5707963267948968]
julia> typeof(ans)
ValidatedNumerics.Interval{Float64}
```

I do not know where this is induced, but my way of thinking is that, if you use `@biginterval` is because you mean it, so the answer should be a `Interval{BigFloat}`. This is probably related to `convert(Interval,x)`, commented elsewhere.

lbenet:
In general I agree with merging. Yet, I don't think that `convert(Interval,x::Real)` should return an `Interval{Float64}` by default. I added a comment, I guess related to this, in one of the tests.

test/interval_tests/trig_tests.jl
dpsanders: This is because of a fallback to a Julia definition `atan2(x, y)` when `x` and `y` have different types, which passes `x` and `y` through `float`. I think this only worked before because the `BigFloat` precision was set to 53; the type was previously `Interval{Float64}` as well, but we didn't realise!

If you want to insist on `@biginterval`s, presumably you want it everywhere, in which case this does the correct thing:

```
julia> atan2(@biginterval(0.0, 1.0), @biginterval(0.0))
[1.570796326794896619231321691639751442098584699687552910487472296153908203143099, 1.570796326794896619231321691639751442098584699687552910487472296153908203143116]₂₅₆
```

dpsanders:
> In general I agree with merging. Yet, I don't think that convert(Interval,x::Real) should return an Interval{Float64} by default. I added a comment, I guess related to this, in one of the tests.

I just left this as it was previously, e.g. in v0.2.0:
```
julia> convert(Interval, 1//3)
[0.3333333333333333, 0.33333333333333337]
```

The behaviour you suggest is already provided by the `Interval` constructor:
```
julia> Interval(1//3)
[1//3, 1//3]

julia> Interval(big(1.))
[1.0000000000000000, 1.0000000000000000]₅₃
```
I actually think that it is too surprising if `convert(Interval, x)` also works like that, since the type of the output will depend on the type of the input. 

Fot that reason I think we should remove `convert(Interval, x)` entirely and only allow `convert(Interval{T}, x)` for a specific `T`.

src/intervals/hyperbolic_functions.jl
dpsanders: Thanks, fixed.

src/intervals/rounding.jl
dpsanders: Thanks, fixed.

dpsanders:
One last thing that bothers me is the behaviour converting to a `Interval{Rational{T}}`.
I don't see why we would want to permit things like
```
julia> convert(Interval{Rational{Int}}, pi)
[165707065//52746197, 165707065//52746197]
```
i.e. converting floats to rationals. The argument for keeping rational intervals is that they are exact (although the rationals are already exact, so I'm not sure what we gain by having intervals). Here we have converted to a thin interval that does not contain the true value of `pi`.

I think we should just disallow conversions from floats to rational intervals. You can always do it explicitly yourself using `rationalize` if you want.

lbenet:
Two comments about your comments:

> I actually think that it is too surprising if convert(Interval, x) also works like that, since the type of the output will depend on the type of the input.

That's ok: multiple dispatch!

> Fot that reason I think we should remove convert(Interval, x) entirely and only allow convert(Interval{T}, x) for a specific T.

I think `convert(Interval, x::Real) = Interval{typeof(x)}(x)` would be ok.

lbenet:
Is this ready to be merged? The only point that still remains is related to `convert(Interval,x::Real)`...

dpsanders:
Yes it's ready to be merged apart from deciding about convert(Interval).

lbenet:
So, do we keep as it is now (I'd rather not), do we erase it  or do we change it as I proposed above?

dpsanders:
We can change it as you proposed. Do you want to do it or shall I?

dpsanders:
I'll do it later.

dpsanders:
I tried 
```
convert{T<:Real}(::Type{Interval}, x::T) = convert(Interval{T}, x)
```
but then what should `convert(Interval, x::Int64)` give? 
Which type should automatically be chosen? There is no reasonable answer.

This seems to me to show that the `convert(Interval, x)` really is not well defined.

lbenet:
So let's go with your proposal of simply deleting it, and then merging.

dpsanders:
Passing on 0.4, so merging! Thanks for the comments and discussion.

Issue #93, opened 2016-03-19
Break up long test suites


dpsanders:
To deal with the (hopefully temporary) issue https://github.com/JuliaLang/julia/issues/15346, we should break up the long test suites (say, longer than 100 tests) into sub-parts.

lbenet:
I've checked where the tests are hanging: locally, it is in the pow tests of the ITF1788 suite, though travis hangs elsewhere (after sqrt). My point is that our own tests run nicely. So perhaps it is enough to add an `if` requiring v0.4 in order to check the ITF1788 test suite, so v0.5 will not check them. Eventually, when v0.5 gets stabilized, we should take away that if and test everything.

How do you like this idea?

dpsanders:
Great idea! I'll do that.

dpsanders:
@cpierard: Would you like to try doing this?

dpsanders:
In particular this file: https://github.com/dpsanders/ValidatedNumerics.jl/blob/master/test/ITF1788_tests/libieeep1788_tests_elem.jl

PR #100, opened 2016-04-02
Refactor files to remove small files and simplify file names

dpsanders:
- Simplify file names
- Moved macros from rounding.jl into macros.jl, and the non-macro stuff from rounding.jl into precision.jl
- Removed rationalize.jl that is no longer used
- Move short printing stuff into intervals.jl


src/intervals/intervals.jl
```diff
@@ -38,16 +38,46 @@ eltype{T<:Real}(x::Interval{T}) = T
...
-include("rationalize.jl")
-include("parameters.jl")
-include("rounding.jl")
-include("macro_definitions.jl")
-include("conversion_promotion.jl")
+include("special.jl")
+include("macros.jl")
+include("conversion.jl")
 include("arithmetic.jl")
 include("precision.jl")
```
lbenet: The file "precision.jl" contains, among others, the `IntervalParameters` type and other rounding functions. Though it seems to create no problems for the tests, for sanity I think it is better to include this file *before* any actual use of parameters or rounding stuff. In particular, include it before "arithmetic.jl".

lbenet:
I think I can merge it once my sole comment is implemented.

lbenet:
It may be worth to implement the idea in #93 before merging this.

PR #101, opened 2016-04-02
Replace our autodiff routines with the ForwardDiff.jl package

dpsanders:
It's pretty amazing that there's so little to do here!


lbenet:
Beautiful !

Ready to merge, once #100 and #102 are ready.

PR #102, opened 2016-04-02
Refactor with_interval_precision -> setprecision etc.

dpsanders:
Changes:
- `set_interval_precision(x)` -> `setprecision(Interval, x)`.
- `get_interval_precision()` -> `precision(Interval)`
- `set_interval_rounding(x)` -> `setrounding(Interval, x)`
- `get_interval_rounding()` -> `rounding(Interval)`

in accordance with https://github.com/JuliaLang/julia/pull/13232
i.e. don't invent a new function, just add methods to the (now) standard functions.

### DIFF
--- a/src/ValidatedNumerics.jl
+++ b/src/ValidatedNumerics.jl
@@ -8,10 +8,6 @@ using CRlibm
 using Compat
 using FixedSizeArrays
 
-
-setrounding(BigFloat, RoundNearest)
-setrounding(Float64, RoundNearest)
-
 import Base:
     +, -, *, /, //, fma,
     <, >, ==, !=, ⊆, ^, <=,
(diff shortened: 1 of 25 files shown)


lbenet:
I'll consider the last two commits, since the rest is contained in #100.

src/ValidatedNumerics.jl
```diff
@@ -26,25 +22,31 @@ import Base:
...
     expm1, log1p,
-    isfinite, isnan
+    isfinite, isnan,
+    precision
+
+if VERSION >= v"0.5.0-dev+1182"
+    import Base: rounding, setrounding, setprecision
+else
+    import Compat: rounding, setrounding, setprecision
+end
```
lbenet: I don't quite understand why this is written like this. 

Isn't it enough to add `@compat` when it is needed, avoiding this `if`?

src/ValidatedNumerics.jl
```diff
@@ -26,25 +22,31 @@ import Base:
...
-    set_interval_precision, get_interval_precision,
-    with_interval_precision,
     parameters, eps, dist, roughly,
     pi_interval,
     midpoint_radius, interval_from_midpoint_radius,
     RoundTiesToEven, RoundTiesToAway,
     cancelminus, cancelplus, isunbounded,
-    .., @I_str
+    .., @I_str,
+    precision, setprecision, setrounding, rounding
```
lbenet: Do we have to export `precision`, `setprecision`, `setrounding`, `rounding`? I thought we were simply adding new methods?

src/ValidatedNumerics.jl
```diff
@@ -60,8 +62,11 @@ export
     find_roots_midpoint
 
 function __init__()
```
lbenet: Nice!

lbenet:
Probably a good idea to rebase before merging (once #100 is merged).

src/ValidatedNumerics.jl
dpsanders: I'm not sure if we could use `@compat`, but since these commands, in particular `setprecision`, occur a lot, I would not like to do so.

This is certainly ugly, but I'm not sure if there's a simple way to improve it.

src/ValidatedNumerics.jl
dpsanders: I believe we have to reexport them if they were imported from `Compat` (rather than `Base`), so that the user doesn't have to do `using Compat` in order to use them.

lbenet:
The problem with the tests (0.5) may be solved by rebasing...

PR #103, opened 2016-04-05
Temporarily disable ITF1788 tests on 0.5

dpsanders:


dpsanders:
Tests now pass on 0.4 and 0.5, so merging, with the caveat that we should periodically check if the situation has been resolved!

PR #112, opened 2016-04-25
Add decorated intervals

lbenet:
This PR implements the decoration system, and it is tested against the ITF1788 suite. (The tests with Julia 0.5 were hanging locally on my machine.) I needed to rebase from the branch "decorations" to include a bunch of changes (since version 0.2). Still missing is to extend our own tests. I think this PR is ready for reviewing and corrections.


lbenet:
Also, the documentation is missing...

dpsanders:
Great work, thanks! Let me look through it carefully.

lbenet:
As you'll see, it is heavily based on yours, extending it whenever was needed.

dpsanders:
It's pretty amazing that the ITF1788 tests pass! Really nice work extending what I had done. 

Since all the tests pass, maybe we should just merge this and then iterate on it? Or what else do you think needs to be done?

src/decorations/functions.jl
```diff
@@ -0,0 +1,315 @@
...
+abs{T}(xx::DecoratedInterval{T}) =
+    DecoratedInterval(abs(interval(xx)), decoration(xx))
+
+
+## Other (cancel and set) functions
+other_functions = ( :cancelplus, :cancelminus, :intersect, :hull, :union )
+
+for f in other_functions
+    @eval $(f){T}(xx::DecoratedInterval{T}, yy::DecoratedInterval{T}) =
+        DecoratedInterval($(f)(interval(xx), interval(yy)), trv)
```
dpsanders: I don't think just putting `trv` is correct. Shouldn't it be the `decay` of the others? Maybe not for `hull` (which is sort of a discontinous operation), but for the others (which seem to be in some sense "continuous" operations). Or is this discussed in the Standard?

src/decorations/decorations.jl
```diff
@@ -0,0 +1,90 @@
...
+a flag that records the status of the interval when thought of as the result
+of a previously executed sequence of functions acting on an initial interval.
+"""
+type DecoratedInterval{T<:Real} <: AbstractInterval
+    interval::Interval{T}
+    decoration::DECORATION
+
+    function DecoratedInterval(I::Interval, d::DECORATION)
+        dd = decoration(I)
+        Int(dd) < 2 && return new(I, dd)
```
dpsanders: Use symbolic values here, not numeric: `dd <= trv`

src/decorations/decorations.jl
```diff
@@ -0,0 +1,90 @@
...
+DecoratedInterval{T<:Real}(a::T) = DecoratedInterval(Interval(a,a))
+DecoratedInterval(a::Tuple) = DecoratedInterval(Interval(a...))
+DecoratedInterval{T<:Real, S<:Real}(a::T, b::S) = DecoratedInterval(Interval(a,b))
+
+interval(x::DecoratedInterval) = x.interval
+decoration(x::DecoratedInterval) = x.decoration
+
+# Automatic decorations for an interval
+decoration(I::Interval) =
+    ifelse( isnai(I), ill,           # nai()
```
dpsanders: The nested `ifelse` are hard to read.
Replace by `isnai(I) && ill` etc.

src/ValidatedNumerics.jl
```diff
@@ -55,6 +57,12 @@ end
 export
     IntervalBox, @intervalbox
 
+## Decorations
+export
+    @decorated,
+    interval, decoration, DecoratedInterval,
```
dpsanders: Maybe `interval` is not the best name (my fault, I know). `interval_part`? Does this need to be exported?

src/decorations/decorations.jl
```diff
@@ -0,0 +1,90 @@
...
+DecoratedInterval(I::Interval) = DecoratedInterval(I, decoration(I))
+DecoratedInterval{T<:Real}(a::T, b::T) = DecoratedInterval(Interval(a,b))
+DecoratedInterval{T<:Real}(a::T) = DecoratedInterval(Interval(a,a))
+DecoratedInterval(a::Tuple) = DecoratedInterval(Interval(a...))
+DecoratedInterval{T<:Real, S<:Real}(a::T, b::S) = DecoratedInterval(Interval(a,b))
+
+interval(x::DecoratedInterval) = x.interval
+decoration(x::DecoratedInterval) = x.decoration
+
+# Automatic decorations for an interval
```
dpsanders: It's clever to use the same name `decoration` for this!

src/decorations/functions.jl
lbenet: It is discussed in the standard; there, it is recommended that the user adapts it by changing the decoration. I'll add this (and other) docstrings explaining this.

src/decorations/decorations.jl
lbenet: I'll implement it as you suggest.

src/decorations/decorations.jl
lbenet: I agree; I don't know what was I thinking...

src/ValidatedNumerics.jl
lbenet: I'll change it to `interval_part`. I think it is ok to simply export it; it is actually used in the tests.

src/decorations/decorations.jl
lbenet: Thanks!

lbenet:
I've just pushed a new commit following your suggestion; it includes some docstrings. (I hope to push a new one with more docstrings; just give me a few days.)

lbenet:
Incidentally, can you take a look on `@decorated`; I don't think it works...

lbenet:
Also, other tests should be added, covering decorated intervals with extended precision.

dpsanders:
I think we should merge this now.

lbenet:
Fine with me.

Issue #39, opened 2015-06-22
Add option to display intervals as `3 ± 0.1`


dpsanders:
Implement @lbenet's suggestion in a comment on #35, we should implement a display option to show the interval `[2.9, 3.1]` as `3.0 ± 0.1`, using `midpoint_radius`.

This could be a global option in the `IntervalParams` type, and/or a keyword argument to `show`.

lbenet:
Just to be very clear: I think this is worth specially for the midpoint_radius representation, not necessarily in general.

dpsanders:
Fixed by #115.

PR #116, opened 2016-04-27
Add \pm notation for creating interval in midpoint--radius representation

dpsanders:
```julia
julia> 3 ± 0.5
[2.5, 3.5]
```


lbenet:
This addresses #39 (and related to #115).

PR #118, opened 2016-05-08
Rewrite decoration docs with examples

dpsanders:


docs/decorations.md
```diff
@@ -1,47 +1,145 @@
-# Decorations:
-
-Decorations are flags or labels attached to intervals to indicate the status of a given interval as the product of a function evaluation on a given initial interval.
-
-The possible decorations and their ordering are as follows (pg. 46):
+[I 160508 10:04:49 handlers:132] Browser Connected: http://127.0.0.1:8000/multid<script type="text/x-mathjax-config">
```
lbenet: All these lines (related to MathJax) will not be needed once we use Documenter.jl; I suggest to get rid of them.

lbenet:
Is this ready to be merged? I think it is...

docs/decorations.md
dpsanders: For the moment they are required to display mathematics. We will certainly be happy to get rid of them once we have a better solution!

PR #125, opened 2016-05-11
REQUIRE latest tagged version of ForwardDiff

dpsanders:


dpsanders:
The latest tagged version of ForwardDiff removes some redefinition and ambiguity warnings on Julia 0.5.

Issue #124, opened 2016-05-10
Make decorated intervals the default


dpsanders:
Replace:
- `Interval`-> `BareInterval`
- `DecoratedInterval` -> `Interval`

Also make multi-dimensional `IntervalBox`es work with decorated intervals.

lbenet:
I actually like better the way it is now: `Interval` is the short form which you use for exploration; then you do things carefully with `DecoratedInterval`.

As for `IntervalBox`, I agree it should work with decorated intervals.

dpsanders:
There is no need to make a difference between them, though -- you just leave off the output of decorations (which is, in any case, the default), and everything just works ;)

dpsanders:
In other words, you get decorations for free, which is why they should be the default.

lbenet:
I agree that `Interval` is already contained in `DecoratedInterval`. Yet, the standard makes a distinction among them, being the bare intervals more basic, so I am in favor of keeping them separately.

dpsanders:
Yes, they have to be separate for the standard.
I am suggesting that we rename the current `Interval` to `BareInterval`, and rename the current `DecoratedInterval` to the new `Interval`, which will be the new default.

A "standard user" will not notice a single difference after this change, but in fact the same calculation, written as before, now provides more information in the decorations when the user goes looking for them.

dpsanders:
This is proving to be more difficult than I thought. The idea, though, is that this change should be *completely transparent* for the user: `@interval` should now just make a decorated interval instead of a bare interval, and everything else should then just work...

lbenet:
I'm really not in favor of this change and #126. For a standard user, the propagation of decorations means a (marginal) penalty on performance. If, by some reason, the calculation is long enough, this will eventually be important. I don't think it represents any value for that standard user. 

Also, as it is now, it is simpler to check whether things are working or not, since we only care about the `Interval` part, and decorations are, precisely, extensions to that. And, decorations are tricky and in some cases should be reset manually.

In a way, this change seems similar to using by default extended precision (256 bits), which indeed contains `Float64`, so the user can always go back. My point of view is that you should opt for `BigFloat` if you really need it, but not by default.

dpsanders:
I certainly see your point and your argument.
@oheim had recommended making decorated intervals the default in a previous discussion.

[...]

PR #126, opened 2016-05-12
Make Interval refer to decorated intervals

dpsanders:


dpsanders:
Ouch, this will be more difficult than I thought...

lbenet:
Ups, I wrote my opinion in #124, and not here....

dpsanders:
OK, closing this. But if we decide to do it eventually, it is not *too* difficult. (Though I think it's more subtle than what I was doing here.)

Issue #127, opened 2016-05-12
`@interval(3,2)` not throwing an ArgumentError


lbenet:
Rereading [this comment](https://github.com/dpsanders/ValidatedNumerics.jl/pull/55#issuecomment-139120253), I just realized that `@interval` and `@decorated` are not behaving as they should. For example:
```julia
julia> using ValidatedNumerics

julia> displaymode(format=:full, decorations=true)

julia> @interval(3,2)  # should yields an error!
Interval(2.0, 3.0)

julia> @decorated(3,2) # should yield nai with ill-decoration
DecoratedInterval(Interval(2.0, 3.0), com)
```

This is in contrast with the use of `Interval` or `DecoratedInterval` directly.
```julia
julia> Interval(3,2)
ERROR: ArgumentError: Must have a ≤ b to construct Interval(a, b).
 in call at /Users/benet/.julia/v0.4/ValidatedNumerics/src/intervals/intervals.jl:19
 in call at /Users/benet/.julia/v0.4/ValidatedNumerics/src/intervals/intervals.jl:36

julia> DecoratedInterval(3,2)
ERROR: ArgumentError: Must have a ≤ b to construct Interval(a, b).
 in call at /Users/benet/.julia/v0.4/ValidatedNumerics/src/intervals/intervals.jl:19
 in call at /Users/benet/.julia/v0.4/ValidatedNumerics/src/decorations/intervals.jl:51
```

The result is because the use of `hull` in the macros. I have no idea how to make that this behaves correctly.

dpsanders:
Good point. Instead of hull we can just use a.lo and b.hi.   I'll do this later when I have a chance.

lbenet:
I just played with your idea, but got stock elsewhere...

dpsanders:
Where is it getting stuck? I am not at a computer right now.

lbenet:
I'm adapting the tests; the problem was the way I was modifying the macro.

dpsanders:
Closed by #128.

PR #128, opened 2016-05-12
Solve issue #127 and add more tests

lbenet:


lbenet:
Note that `DecoratedInterval(2//1,1//2)` throws an `ArgumentError`, since there are no `NaN`s for `Rational{T}`.

src/decorations/intervals.jl
```diff
@@ -92,9 +98,9 @@ macro decorated(ex...)
     local x
 
     if length(ex) == 1
-        x = :(@interval($(esc(ex[1]))))
+        x = :($(esc(ex[1])))
```
dpsanders: Won't this fail for an argument like pi? (not checked)

lbenet: You may be right, though tests pass. I'll revert that change, because it is actually not needed. Thanks!

lbenet:
Incidentally, tests in julia 0.5 simply get stuck in the decoration tests! I'm trying to understand what's going on there...

lbenet:
In Julia 0.5, `DecoratedInterval(1//1, 1//10)` yields a weird error, with the first few lines being:

```julia
julia> DecoratedInterval(1//1, 1//10,com)
"WARNING: An error occured during inference. Type inference is now partially disabled."
OverflowError()
...
```

I am disabling the last two tests, which are somehow related...

lbenet:
Tests are passing.... I don't understand codecov/patch

dpsanders:
LGTM, thanks! I think codecov/patch refers to the fact that the coverage *of your new patch* (i.e. of the changes in this PR) has coverage which is lower than that of the project overall.

PR #143, opened 2016-05-27
Set operations and setdiff

dpsanders:


dpsanders:
Merging and tagging 0.4.1, since the current tagged version breaks with a new `FixedSizeArrays`.

PR #146, opened 2016-06-04
Improve display of IntervalBox with full display option

dpsanders:


dpsanders:
Tests pass on 0.4 but hang on 0.5. Merging.

Issue #147, opened 2016-06-21
Breaking changes in ForwardDiff


lbenet:
ForwardDiff was recently upgraded and version 0.2 introduces several changes that break our root_finding code. See [this](http://www.juliadiff.org/ForwardDiff.jl/upgrade.html).

lbenet:
This is the error
```julia
ERROR: LoadError: LoadError: UndefVarError: derivative not defined
 in include at /usr/local/julia/julia-57d08348c5/lib/julia/sys.dylib
 in include_from_node1 at /usr/local/julia/julia-57d08348c5/lib/julia/sys.dylib
 in include at /usr/local/julia/julia-57d08348c5/lib/julia/sys.dylib
 in include_from_node1 at /usr/local/julia/julia-57d08348c5/lib/julia/sys.dylib
 in process_options at /usr/local/julia/julia-57d08348c5/lib/julia/sys.dylib
 in _start at /usr/local/julia/julia-57d08348c5/lib/julia/sys.dylib
while loading /Users/benet/.julia/v0.4/ValidatedNumerics/test/root_finding_tests/root_finding.jl, in expression starting on line 4
while loading /Users/benet/.julia/v0.4/ValidatedNumerics/test/runtests.jl, in expression starting on line 22
```

dpsanders:
Done in #148

PR #148, opened 2016-06-21
Deal with updates in derivative (ForwardDiff v0.2)

lbenet:
@dpsanders Updates the syntax changed by ForwardDiff v0.2 in `derivative` which were breaking the root-finding tests; see #147.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,4 +2,4 @@ julia 0.4
 CRlibm 0.2.2
 Compat 0.7.11
 FixedSizeArrays 0.1.2
-ForwardDiff 0.1.8
+ForwardDiff 0.2.0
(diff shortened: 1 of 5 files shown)


dpsanders:
LGTM, thanks!

dpsanders:
LGTM, thanks!
Does this work *only* with ForwardDiff 0.2? Should we update that in REQUIRE?

lbenet:
Yes, the changes are actually motivated by changes in v0.2.0 of ForwardDiff. I guess we should include this in REQUIRE, since i don't think this is compatible with earlier version; can you check this? This change then will eventually force us to tag a minor version.

dpsanders:
OK, could you add the change to REQUIRE then?

lbenet:
Done.

dpsanders:
Thanks!

tkelman:
Would you mind making a tag that includes this?

PR #149, opened 2016-06-22
Solve promotion warnings triggered by ForwardDiff

lbenet:
This removes some warnings due to the recent update of ForwardDiff. The `promote_rule`s introduced produce
```julia
julia> promote(ForwardDiff.Dual(2,1), Interval(1,2))
(Dual([2, 2],[1, 1]),Dual([1, 2],[0, 0]))
```
No tests have been included yet.


dpsanders:
Could you add a couple of tests, please?

I wonder if we should move the root finding stuff into a submodule?

lbenet:
Please, don't merge it yet. I have included other tests, which I will push soon.

lbenet:
I just pushed the tests... (I just didn't want that travis ends and this PR is merged.)

dpsanders:
Since `ForwardDiff` is used only for root finding, I think we should segregate out the uses of it, so put these tests in the `root_finding_tests` directory where I put the one I added.

lbenet:
There is actually only (new) test that is related to ForwardDiff; the other should in principle increase the coverage. Shall I move it to the place you propose?

dpsanders:
Well I mean that I think we should separate out both the usage of ForwardDiff and its testing, and maybe make an actual RootFinding submodule of ValidatedNumerics.

dpsanders:
Merging for now, thanks!

Issue #152, opened 2016-06-26
ArgumentError of cos(Interval(-pi/2, 3pi/2))


kriukov:
I am running into a problem with the cos() function. When I try to do

cos(Interval(-pi/2, 3pi/2))

or

cos(Interval(-1.5707963267948966, 4.71238898038469))

I get an ArgumentError:

ERROR: ArgumentError: Must have a <= b to construct Interval(a, b)

(with a, b = 6.123233995736765e-17, -1.8369701987210297e-16).

lbenet:
I understand the problem; I'll try to look onto it tomorrow.

dpsanders:
This is also wrong:
```
julia> sin(Interval(-pi/2, 3pi/2))
[-1, -0.999999]
```

dpsanders:
Fix proposed in #155 . These should return `[-1,1]`, correct?

lbenet:
Just a short remark:

```julia
julia> cos(Interval(-pi/2, 3pi/2))
ERROR: ArgumentError: Must have a ≤ b to construct Interval(a, b).
 in call at /Users/benet/.julia/v0.4/ValidatedNumerics/src/intervals/intervals.jl:19
 in cos at /Users/benet/.julia/v0.4/ValidatedNumerics/src/intervals/trigonometric.jl:86

julia> cos(@interval(-pi/2, 3pi/2))
[-1, 1]
```

The point of this is that the use of `Interval` with -pi/2 and 3pi/2 is not correct to build an Interval that contains *both* -pi/2 and 3pi/2:
```julia
julia> displaymode(format=:full)

julia> Interval(-pi/2, 3pi/2)
Interval(-1.5707963267948966, 4.71238898038469)

julia> @interval(-pi/2, 3pi/2)
Interval(-1.5707963267948968, 4.712388980384691)
```

In any case, something should be corrected because, despite of the issue of using `Interval` or `@interval`, the construction of the result should (automatically) work.

dpsanders:
I agree that `@interval` is the semantically correct thing to use here, but 
`cos(Interval(-pi/2, 3pi/2))` is a valid thing to write (it is an interval between two floating point numbers) and should give a (correct) result!

lbenet:
[...]

PR #155, opened 2016-06-28
Fix sin and cos for values near quadrant boundaries

dpsanders:


lbenet:
Note that in line 32 of `src/intervals/trigonometric.jl` (caring only about `sin` for the time being), it is checked if the diam of the interval in question is larger than 2pi, actually, larger than the lower bound for the (corrected rounded) interval corresponding to 2pi.

lbenet:
i don't think your solution is actually correct; see my comment: https://github.com/dpsanders/ValidatedNumerics.jl/issues/152#issuecomment-229136158. We have to make sure that the lower and upper limits of the produced intervals are ordered.

dpsanders:
Why don't you think my solution is correct? It turns out that this possibility was overlooked by the line after my change, which immediately takes mod 4 of the quadrants, but for this input the min and max quadrants differ by 5.

lbenet:
Internally, the following is happening when `cos(Interval(-pi/2,3pi/2)` is evaluated: 

`Interval(-pi/2, 3pi/2)` has a width (due to Float64 rounding) which is smaller than `ValidatedNumerics.two_pi(Float64).lo`, so it does not return the interval [-1,1]. This is either because `-pi/2` is rounded up, or `3pi/2` is rounded down, or both (actually, the latter is true).

The first evaluation of `lo_quadrant` yields -2 for `-pi/2` and `hi_quadrant` yields 3; so `lo_quadrant` is **not** reflecting that `-pi/2` is rounded up(!), whereas `hi_quadrant` is reflecting that `3pi/2` is rounded down. The second evaluation yields `lo_quadrant` equal 2, and `hi_quadrant` equal 3.

With these values the first `elseif` case is met, which returns `Interval(cos(a.lo, RoundDown), cos(a.hi, RoundUp))`. The lower limit of the returned interval is just above zero, while the upper limit is just below zero; hence the error.

I think the problem might be in `find_quadrants`, rather than in `cos` or `sin`.

dpsanders:
The `find_quadrants` function is deliberately pessimistic, so indeed gives the "wrong" answer for `-pi/2`; I don't see how to get around this, though.

dpsanders:
Both `lo_quadrant` and `hi_quadrant` reflect the fact that the values are "right on" quadrant boundaries:
```
julia> h = @interval(pi)/2
[1.57079, 1.5708]

julia> pi/2 / h
[0.999999, 1.00001]

julia> 3pi/2 / h
[2.99999, 3.00001]
 ```

This division by the *interval*  `@interval(pi)/2` [equivalent to `@interval(pi/2)`] is the only way I can see to do this calculation correctly.

**Comment by lbenet** [comment]
Indeed, `find_quadrants` is pessimistic. Yet, it should be consistent with `diam`, which is not in this case. Your solution seems to be correct indeed, sorry for my previous statement. It is a work around for this inconsistency. But I wonder if such a problem will not pop out again elsewhere.

The fact that `cos(Interval(-pi/2,3pi/2))` and `cos(@interval(-pi/2,3pi/2))` yield different results lies in the test `diam(a) > two_pi(T).lo`. I wonder whether we should consider `2pi` (or a proper conversion) instead of `two_pi(t).lo`...

**Comment by dpsanders** [comment]
`two_pi(T).lo` is equal to `2pi` for `Float64`.

In fact, replacing the test by `diam(a) >= two_pi(T).lo` would fix the result for `(-pi/2, 3pi/2)`, but feels wrong.

I'm not sure what you mean when you say that `find_quadrants` should be consistent with `diam`? 
`find_quadrants` uses divides by an interval version of `pi/2`, which gives correct answers but a slight (harmless) overestimation in this particular case.

**Comment by lbenet** [comment]
My comment was related precisely to the fact that, for both `Interval(-pi/2,3pi/2)` and `@interval(-pi/2,3pi/2)`, `find_quadrants` yield the same result, but the check of `diam` yield something different.

I also noticed that replacing the test could fix the result, but just as you stated, it feels wrong, or at least loose.

**Comment by dpsanders** [comment]
Ah yes, I see what you mean. We can simply remove the `diam` check now.

**Comment by dpsanders** [comment]
Well, we could do that, but then we would be doing additional, unnecessary calculations of the quadrants. I think we can just leave it like this?

**Comment by lbenet** [comment]
Yes, as it is now it works.

What about `tan`? Do we need something similar?

**Comment by lbenet** [comment]
Currently we get
```julia
julia> tan(Interval(-pi/2,pi/2))
[-1.63313e+16, 1.63313e+16]

julia> tan(@interval(-pi/2,pi/2))
[-∞, ∞]
```

lbenet:
I'm not sure, but I think `tan(Interval(-pi/2,pi/2))` is part of the ITF1788 test suite.

dpsanders:
`tan` is correct.
Do you have any objections to merging?

dpsanders:
Since the floats in `(-pi/2, pi/2)` do not contain the discontinuity, but `@interval` rounds these past the discontinuity.

lbenet:
No, no objections.

lbenet:
Merged

PR #156, opened 2016-06-30
Make unique and filter results from Newton

dpsanders:


lbenet:
Nice! Does this work with multiple roots?

dpsanders:
Yes!:
```

julia> f(x) = (x-1)*(x^2-2)^3*(x^3-2)^4; newton(f, -5..5.1, maxlevel=100)
4-element Array{ValidatedNumerics.RootFinding.Root{Float64},1}:
 ValidatedNumerics.RootFinding.Root{Float64}([-1.41422, -1.41421],:unknown)
 ValidatedNumerics.RootFinding.Root{Float64}([1, 1],:unique)
 ValidatedNumerics.RootFinding.Root{Float64}([1.25992, 1.25993],:unknown)
 ValidatedNumerics.RootFinding.Root{Float64}([1.41421, 1.41422],:unknown)
```

lbenet:
Very nice!!

lbenet:
I think you should merge this right away!

src/root_finding/krawczyk.jl
```diff
@@ -79,18 +79,20 @@ function krawczyk{T}(f::Function, f_prime::Function, x::Interval{T}, level::Int=
...
 
-    isthin(x) && return [(x, :unknown)]
+    isthin(x) && return [Root(x, :unknown)]
 
     # bisecting
     roots = vcat(
         krawczyk(f, f_prime, Interval(x.lo, m), level+1,
                  tolerance=tolerance, debug=debug, maxlevel=maxlevel),
-        krawczyk(f, f_prime, Interval(nextfloat(m), x.hi), level+1,
+        krawczyk(f, f_prime, Interval(m, x.hi), level+1,
```
lbenet: I recall that we deliberately decided to use `nextfloat(m)` to make sure that the two intervals considered were somehow separated. (Same comment for newton.jl.) Can you explain the reason behind the change?

dpsanders:
Tests are failing on 0.5 due to some recent upstream breakage, apparently in base Julia.
We could merge anyway though?

src/root_finding/krawczyk.jl
dpsanders: I think that the idea of adding `nextfloat` was a mistake. Conceptually, it is incorrect since it misses out the interval between the two floats (compare unums). From a practical point of view, the current version is better since it allows to check for a non-empty intersection to see if the two intervals are neighbours.

lbenet:
Tests pass in 0.4, and the problems seem to me to be related to FixedSizeArrays and appear in tests related to `IntervalBox`. If you think it will take a long time to solve this, I agree to merge it.

dpsanders:
Since it's not clear how long that might take to fix, I'm going to merge.

Issue #165, opened 2016-08-10
Julia rounding problem


dpsanders:
https://github.com/JuliaLang/julia/issues/17926

This is very bad news for ValidatedNumerics.jl, although until now there doesn't seem to be an issue with our tests? cc @lbenet

lbenet:
For v0.5-dev, we disabled testing ITF-1788 in b9e61b6. Yet, running them locally I see no problem, though they still take too long... I just pushed #166; let's see how travis reacts...

dpsanders:
We do not seem to have an issue with this Julia problem, since we don't do rounding with operations on literals.

PR #166, opened 2016-08-11
Reenable tests for ITF-1788 for v0.5

lbenet:
These tests were disabled a long back, simply because they were
taking too long to complete, so travis was complaning.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -24,9 +24,7 @@ include("root_finding_tests/root_finding.jl")
 
 # ITF1788 tests
 
-if VERSION < v"0.5-dev"
-    include("ITF1788_tests/ITF1788_tests.jl")
-end
+include("ITF1788_tests/ITF1788_tests.jl")
 
 FactCheck.exitstatus()
 


dpsanders:
Are they fast again now?

lbenet:
Tests pass (check travis); in v0.5, they are still running (over 11 min)...

lbenet:
In my machine, tests complete (v0.5) in a little less than 10 min. To compare, the tests complete in v0.4.7-pre+1 in 1 min and 35 secs!!! Huge performance regression!

lbenet:
Travis completed in 16 min!!! I think this should be merged.

Still, the question in #165 is unanswered...

dpsanders:
Yes, go ahead and merge.

I think the answer is that we never do the rounding with constants, only
with variables (a.lo etc).

dpsanders:
Thanks!

PallHaraldsson:
[I think I mentioned somewhere (can find) and got the answer that this is not fully conforming to ITF-1788]

Any idea why so much slower on 0.5? It's not helping, but are lots more tests needed?

lbenet:
(I don't find it either, but there are some things that we have not implemented yet; I think reverse mode functions is one of them.)

Regarding the question on performance, we have also noticed this, but don't quite know what is the problem. Probably profiling it carefully should give some answers. I have read in the mailing lists that other packages are also showing this kind of behavior, so I guess the best is to follow what they are doing.

PR #177, opened 2016-11-21
Proper display for Interval{Rational{T}}

lbenet:
Currently, in master, we get the following:
```julia
julia> using ValidatedNumerics

julia> Interval(1//2, 2//3)
Error showing value of type ValidatedNumerics.Interval{Rational{Int64}}:
ERROR: StackOverflowError:
 in BigInt() at ./gmp.jl:46
 in gcd at ./gmp.jl:255 [inlined]
 in Rational{BigInt}(::BigInt, ::BigInt) at ./rational.jl:9
 in round_string(::Rational{BigInt}, ::Int64, ::RoundingMode{:Down}) at /Users/benet/.julia/v0.5/ValidatedNumerics/src/display.jl:64 (repeats 79993 times)
```

The problem is that it tries to compute the number of significant digits for this interval, which is meaningless for this case.

This PR solves this, so one gets:
```julia
julia> Interval(1//2, 2//3)
[1//2, 2//3]

julia> displaymode(format=:midpoint)

julia> Interval(1//2, 2//3)
7//12 ± 1//12
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ os:
 
 julia:
   - 0.4
+  - 0.5
   - nightly
 
 notifications:
--- a/src/display.jl
+++ b/src/display.jl
@@ -98,7 +98,31 @@ function representation(a::Interval, format=nothing)
     output
 end
 
+function representation{T<:Integer}(a::Interval{Rational{T}},
+    format=nothing)
 
+    if isempty(a)
+        return "∅"
+    end
+
+    if format == nothing
+        format = display_params.format  # default
+    end
+    
+    local output
+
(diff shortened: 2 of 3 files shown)


lbenet:
Some tests (ITF-1788), on v0.6, are not passing... I'll take a look on this later.

dpsanders:
Thanks! Since tests pass on 0.4 and 0.5, maybe we should just merge?

lbenet:
The error in travis is odd; I'm not sure if it is related to some new behaviour in 0.6, or something else. I will restart the tests, and if the same problem persists you can simply merge.

PR #188, opened 2017-01-09
Refactor `@round` to transform `a+b` to `+(a, b, RoundDown)`

dpsanders:
Transforms e.g.
```
@round(a.lo+b.lo, a.hi+b.hi)
```
to
```
Interval(+(a.lo, b.lo, RoundDown), +(a.hi, b.hi, RoundUp))
```

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -155,6 +155,13 @@ end
 
 //(a::Interval, b::Interval) = a / b    # to deal with rationals
 
+function min_ignore_nans(args...)
+    @compat min(Iterators.filter(x->!isnan(x), args)...)
+end
+
+function max_ignore_nans(args...)
+    @compat max(Iterators.filter(x->!isnan(x), args)...)
+end
 
 ## fma: fused multiply-add
 function fma{T}(a::Interval{T}, b::Interval{T}, c::Interval{T})
@@ -177,7 +184,7 @@ function fma{T}(a::Interval{T}, b::Interval{T}, c::Interval{T})
         lo2 = fma(a.lo, b.hi, c.lo)
         lo3 = fma(a.hi, b.lo, c.lo)
         lo4 = fma(a.hi, b.hi, c.lo)
-        min(lo1, lo2, lo3, lo4)
+        min_ignore_nans(lo1, lo2, lo3, lo4)
     end
 
     hi = setrounding(T, RoundUp) do
@@ -185,9 +192,9 @@ function fma{T}(a::Interval{T}, b::Interval{T}, c::Interval{T})
         hi2 = fma(a.lo, b.hi, c.hi)
         hi3 = fma(a.hi, b.lo, c.hi)
         hi4 = fma(a.hi, b.hi, c.hi)
-        max(hi1, hi2, hi3, hi4)
+        max_ignore_nans(hi1, hi2, hi3, hi4)
     end
-    
+
     Interval(lo, hi)
 end
 
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -159,6 +159,12 @@ if VERSION >= v"0.6.0-dev"
     const filter = Iterators.filter
 end
 
+if VERSION < v"0.5"
+    min(x) = x
+    max(x) = x
+end
+
+
 function min_ignore_nans(args...)
     min(filter(x->!isnan(x), args)...)
 end
@@ -166,6 +172,8 @@ end
 function max_ignore_nans(args...)
     max(filter(x->!isnan(x), args)...)
 end
+
+
 
 ## fma: fused multiply-add
 function fma{T}(a::Interval{T}, b::Interval{T}, c::Interval{T})
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -304,7 +304,7 @@ end
 
 function diam{T<:Real}(a::Interval{T})
     isempty(a) && return convert(T, NaN)
-    @setrounding(T, a.hi - a.lo, RoundUp) #cf page 64 of IEEE1788
+    @round_up(a.hi - a.lo) #cf page 64 of IEEE1788
 end
 
 # Should `radius` this yield diam(a)/2? This affects other functions!
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -155,12 +155,16 @@ end
 
 //(a::Interval, b::Interval) = a / b    # to deal with rationals
 
+if VERSION >= v"0.6.0-dev"
+    const filter = Iterators.filter
+end
+
 function min_ignore_nans(args...)
-    @compat min(Iterators.filter(x->!isnan(x), args)...)
+    min(filter(x->!isnan(x), args)...)
 end
 
 function max_ignore_nans(args...)
-    @compat max(Iterators.filter(x->!isnan(x), args)...)
+    max(filter(x->!isnan(x), args)...)
 end
 
 ## fma: fused multiply-add
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -64,12 +64,12 @@ one{T<:Real}(::Type{Interval{T}}) = Interval(one(T))
 
 function +{T<:Real}(a::Interval{T}, b::Interval{T})
     (isempty(a) || isempty(b)) && return emptyinterval(T)
-    @round(T, a.lo + b.lo, a.hi + b.hi)
+    @round(a.lo + b.lo, a.hi + b.hi)
 end
 
 function -{T<:Real}(a::Interval{T}, b::Interval{T})
     (isempty(a) || isempty(b)) && return emptyinterval(T)
-    @round(T, a.lo - b.hi, a.hi - b.lo)
+    @round(a.lo - b.hi, a.hi - b.lo)
 end
 
 
@@ -81,17 +81,17 @@ function *{T<:Real}(a::Interval{T}, b::Interval{T})
     (a == zero(a) || b == zero(b)) && return zero(a)
 
     if b.lo >= zero(T)
-        a.lo >= zero(T) && return @round(T, a.lo*b.lo, a.hi*b.hi)
-        a.hi <= zero(T) && return @round(T, a.lo*b.hi, a.hi*b.lo)
-        return @round(T, a.lo*b.hi, a.hi*b.hi)   # zero(T) ∈ a
+        a.lo >= zero(T) && return @round(a.lo*b.lo, a.hi*b.hi)
+        a.hi <= zero(T) && return @round(a.lo*b.hi, a.hi*b.lo)
+        return @round(a.lo*b.hi, a.hi*b.hi)   # zero(T) ∈ a
     elseif b.hi <= zero(T)
-        a.lo >= zero(T) && return @round(T, a.hi*b.lo, a.lo*b.hi)
-        a.hi <= zero(T) && return @round(T, a.hi*b.hi, a.lo*b.lo)
-        return @round(T, a.hi*b.lo, a.lo*b.lo)   # zero(T) ∈ a
+        a.lo >= zero(T) && return @round(a.hi*b.lo, a.lo*b.hi)
+        a.hi <= zero(T) && return @round(a.hi*b.hi, a.lo*b.lo)
+        return @round(a.hi*b.lo, a.lo*b.lo)   # zero(T) ∈ a
     else
-        a.lo > zero(T) && return @round(T, a.hi*b.lo, a.hi*b.hi)
-        a.hi < zero(T) && return @round(T, a.lo*b.hi, a.lo*b.lo)
-        return @round(T, min(a.lo*b.hi, a.hi*b.lo), max(a.lo*b.lo, a.hi*b.hi))
+        a.lo > zero(T) && return @round(a.hi*b.lo, a.hi*b.hi)
+        a.hi < zero(T) && return @round(a.lo*b.hi, a.lo*b.lo)
+        return @round(min(a.lo*b.hi, a.hi*b.lo), max(a.lo*b.lo, a.hi*b.hi))
     end
 end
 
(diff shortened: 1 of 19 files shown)


lbenet:
Tests are not passing...

dpsanders:
Yes, unfortunately that will require quite a lot of work. But I think this is a good direction to go in.

dpsanders:
Tests are passing on 0.4 and 0.5, and almost on 0.6 with known issues (one in FixedSizeArrays.jl, and the other in Base). cc @lbenet

lbenet:
I'm begining to review this (*huge*) PR... Is it ready to be merged, or the idea is to solve the incompatibilities with Julia v 0.6?

dpsanders:
It's ready to be merged. The only remaining incompatibility is the problem with FixedSizeArrays.jl; I'll see what happens if I change to StaticArrays (in a separate PR).

Sorry about the size of this, but the rounding is so fundamental that it touches almost everything!

dpsanders:
Most of the changes are trivial, though, e.g.
`@round(T, a, b)` loses the `T` to become `@round(a, b)`.

dpsanders:
No problem if you don't feel like reviewing it, though.

src/intervals/set_operations.jl
```diff
@@ -56,18 +56,18 @@ function intersect{T}(a::Interval{T}, b::Interval{T})
...
 
-Returns the "convex hull" of the intervals `a` and `b`, considered as
-(extended) sets of real numbers. That is, the minimum set that contains
-all points in `a` and `b`.
+Returns the "interval hull" of the intervals `a` and `b`, considered as
+(extended) sets of real numbers, i.e. the smallest interval that contains
+all of `a` and `b`.
 """
-hull{T}(a::Interval{T}, b::Interval{T}) = Interval(min(a.lo, b.lo), max(a.hi, b.hi))
+hull{T}(a::Interval{T}, b::Interval{T}) = Interval{T}(min(a.lo, b.lo), max(a.hi, b.hi))
```
lbenet: I just notice that this may be too restrictive, so we need other methods when `eltype(a)` and `eltype(b)` are different. The following yields an error (also in master): `hull(Interval(1//2,3//1), Interval(1,2))`. The same applies to `union`.

src/intervals/trigonometric.jl
```diff
@@ -44,23 +44,25 @@ function sin{T}(a::Interval{T})
...
+        return @round(sin(a.hi), sin(a.lo)) # Interval(sin(a.hi, RoundDown), sin(a.lo, RoundUp))
 
     elseif ( lo_quadrant == 0 || lo_quadrant==3 ) && ( hi_quadrant==1 || hi_quadrant==2 )
-        return Interval(min(sin(a.lo, RoundDown), sin(a.hi, RoundDown)), one(T))
+        return @round(min(sin(a.lo), sin(a.hi)), 1)
+        # Interval(min(sin(a.lo, RoundDown), sin(a.hi, RoundDown)), one(T))
 
     elseif ( lo_quadrant == 1 || lo_quadrant==2 ) && ( hi_quadrant==3 || hi_quadrant==0 )
-        return Interval(-one(T), max(sin(a.lo, RoundUp), sin(a.hi, RoundUp)))
+        return @round(-1, max(sin(a.lo), sin(a.hi)))
```
lbenet: Same comment about conversion; I prefer to use `one(T)` here...

src/intervals/trigonometric.jl
```diff
@@ -86,21 +88,21 @@ function cos{T}(a::Interval{T})
...
-        return Interval(cos(a.lo, RoundDown), cos(a.hi, RoundUp))
+        return @round(cos(a.lo), cos(a.hi))
 
     elseif lo_quadrant == 0 && hi_quadrant==1
-        return Interval(cos(a.hi, RoundDown), cos(a.lo, RoundUp))
+        return @round(cos(a.hi), cos(a.lo))
 
     elseif ( lo_quadrant == 2 || lo_quadrant==3 ) && ( hi_quadrant==0 || hi_quadrant==1 )
-        return Interval(min(cos(a.lo, RoundDown), cos(a.hi, RoundDown)), one(T))
+        return @round(min(cos(a.lo), cos(a.hi)), 1)
```
lbenet: Same comment about use of `1` instead of `one(T)`.

src/intervals/trigonometric.jl
```diff
@@ -86,21 +88,21 @@ function cos{T}(a::Interval{T})
...
 
     elseif ( lo_quadrant == 0 || lo_quadrant==1 ) && ( hi_quadrant==2 || hi_quadrant==3 )
-        return Interval(-one(T), max(cos(a.lo, RoundUp), cos(a.hi, RoundUp)))
+        return @round(-1, max(cos(a.lo), cos(a.hi)))
```
lbenet: Again...

lbenet:
I just noticed that tests are failing again.

lbenet:
The failing tests seem to be related to a change a suggest in `^(Interval{BigFloat},x::AbstractFloat)`... I hope my comment helps.

dpsanders:
Many thanks for the detailed review! I'll work through your comments.

src/intervals/trigonometric.jl
```diff
@@ -27,15 +27,15 @@ end
 function sin{T}(a::Interval{T})
     isempty(a) && return a
 
-    whole_range = Interval(-one(T), one(T))
+    whole_range = Interval{T}(-1, 1)
```
dpsanders: My original motivation for these changes was that I was suspecting that they were introducing a type instability problem (which I doubt, but I just wanted to make sure).

I also think that the `Interval{T}(-1, 1)` is easier to read.

However, now that you pointed it out, I went digging, and found that 
`@edit one(Float64)`
gives the following definition:
```
one{T<:Number}(::Type{T}) = convert(T,1)
```
So it's doing the conversion anyway ;)

Also if you define 
```
f{T}(::Type{T}) = Interval(-one(T), one(T))
g{T}(::Type{T}) = Interval{T}(-1, 1)
```
and do `@code_native f(Float64)` and similarly for `g`, you'll find that there are fewer instructions for `g`. So all round, the new version seems to be a win!

src/intervals/trigonometric.jl
dpsanders: Here the reason is actually different to change this to not use `one(T)`, namely that it was causing problems with the new version of the macro. I now see that these should be solvable; nonetheless, I actually find it much more readable like this, and the same comment about efficiency (that the conversion still happens anyway) applies.

src/intervals/arithmetic.jl
```diff
@@ -165,25 +184,29 @@ function fma{T}(a::Interval{T}, b::Interval{T}, c::Interval{T})
...
+
     end
 
     lo = setrounding(T, RoundDown) do
         lo1 = fma(a.lo, b.lo, c.lo)
         lo2 = fma(a.lo, b.hi, c.lo)
         lo3 = fma(a.hi, b.lo, c.lo)
         lo4 = fma(a.hi, b.hi, c.lo)
-        min(lo1, lo2, lo3, lo4)
+        min_ignore_nans(lo1, lo2, lo3, lo4)
```
dpsanders: That is because that is parsed as `(nai() * 1)..2`. When you write it as `nai() * (1..2)`, it does correctly give `[nan, nan]`.  By the way, I'm not sure why `nan` is output instead of `NaN`?

dpsanders: By the way, maybe it is wrong that `Interval` allows to do `Interval(NaN, 2)` at all?

src/intervals/functions.jl
```diff
@@ -18,68 +18,71 @@ function ^(a::Interval{BigFloat}, n::Integer)
...
+                return @round(a.lo^n, a.hi^n)
             else
-                return @round(T, mag(a)^n, mig(a)^n)
+                return @round(mag(a)^n, mig(a)^n)
             end
         end
     end
 end
 
 function sqr{T<:Real}(a::Interval{T})
```
dpsanders: The difference is almost negligible (a couple of comparisons), and it just duplicates code unnecessarily.

src/intervals/functions.jl
```diff
@@ -18,68 +18,71 @@ function ^(a::Interval{BigFloat}, n::Integer)
...
+    #     return @round(a.lo^2, a.hi^2)
+    #
+    # elseif a.hi ≤ zero(T)
+    #     return @round(a.hi^2, a.lo^2)
+    # end
+    #
+    # return @round(mig(a)^2, mag(a)^2)
 end
 
+^(a::Interval{BigFloat}, x::AbstractFloat) = a^big(x)
```
dpsanders: I don't think it's really necessary. If you want to be careful you should anyway be doing `a^(1//3)` or something like that. I'm actually beginning to wonder if we should be "guessing" the user's intentions anyway. (There was a thread on discourse or somewhere about this recently.)

src/intervals/set_operations.jl
dpsanders: I doubt that in an actual application you would be mixing interval types like that. But we can certainly add `hull(a, b) = hull(promote(a, b)...)`.

src/intervals/trigonometric.jl
lbenet: I partially agree with your argument 😄. One the one hand, you are right, the code is more readable, the conversion is also there, and calling `one` produces more code. Yet, I can't forget the usual example of type instability where the *extra* conversion saves the day in terms of performance. In any case, maybe for `Interval` the whole discussion is somewhat irrelevant, sorry!, in the sense that we `Interval{Int} is converted by using `float()`. So let's leave it as it is now, and forget my comments on this.

src/intervals/arithmetic.jl
lbenet: I agree; I didn't notice the parsing issue before...

src/intervals/functions.jl
lbenet: Yet, our tests are failing right here: the test checks that `x^(1//3) --> x^(1/3)` holds, for `x = biginterval(27)`; the correct answer *must* contain 3. So this is the method that is being called for the RHS. 

Checking this test, `big(1//3)` behaves correctly (renders `BigInt(1)//BigInt(3)`), and the answer contains `3`. However, `big(1/3)` obviously behaves incorrectly (converts the `Float64` representation of `1/3` to `BigFloat`, which is far away from 1/3!).  Incidentally, my proposal is not working either for this.

src/intervals/set_operations.jl
lbenet: I think we should add such a method, as well as the corresponding for `union`. Indeed, it maybe rare in applications...

dpsanders: Fallback added.

src/intervals/trigonometric.jl
dpsanders: You were right to raise the issue (and I knew you would when I made the change ;) ). It led to a useful discussion. But I suggest that we leave the new version.

dpsanders:
I split out the rounding macros into a new file, `rounding_macros.jl`. 
(Actually there is now only one rounding macro.)

dpsanders:
I should say that part of the motivation for this whole change, which was in my head but not necessarily mentioned anywhere, was that this will make it much easier to experiment with different types of rounding: the only place that needs to be changed are now the definitions in `rounding.jl`!

src/intervals/functions.jl
dpsanders: I can imagine the possibility that you would explicitly want to see what happened with an the floating point number "1/3". The alternative is to convert the exponent to an `Interval`.

dpsanders: For now I just removed that test and everything now seems to be passing. We could just record this as an issue and leave it for now.

dpsanders:
Tests are passing on 0.4 and 0.5. I think this is ready to merge!

dpsanders:
The macro documentation is in `rounding_macros.jl` -- see https://github.com/dpsanders/ValidatedNumerics.jl/blob/7e67cbd3cbf3b895a29c8839221f3f9e0bb7ff3e/src/intervals/rounding_macros.jl

Is that what you were looking for? 

Since `1/3` is not the true one third, `x^(1/3)` doesn't need to contain the "true" answer for the true one-third power. 

I see this as analogous to the fact that you can do `Interval(1/3)` if you want to use the floating-point number closest to the true one third, or use `@interval(1/3)` if you want to include the true one third.

Similarly you can do `x^(1/3)` to use the floating-point value `1/3`, or e.g. `x^(1//3)` to use the true real number one third.

dpsanders:
I think we should merge this and can come back and change things later if necessary.

lbenet:
Ok. I'm merging.

dpsanders:
Thanks!

src/intervals/arithmetic.jl
```diff
@@ -155,6 +155,25 @@ end
 
 //(a::Interval, b::Interval) = a / b    # to deal with rationals
 
+if VERSION >= v"0.6.0-dev"
```
tkelman: this won't work for all 0.6-dev versions, should be more specific about version-depedent cutoffs or use `isdefined`

src/intervals/arithmetic.jl
```diff
@@ -155,6 +155,25 @@ end
...
+    const filter = Iterators.filter
+end
+
+if VERSION < v"0.5"
+    min(x) = x
+    max(x) = x
```
tkelman: since you've imported these, this is type piracy and could change the behavior of other unrelated code

src/intervals/rounding.jl
```diff
@@ -0,0 +1,75 @@
+# Define rounded versions of elementary functions
```
tkelman: didn't you have PR's to add these to base?

dpsanders: No, just for `BigFloat` constructors.

src/intervals/arithmetic.jl
dpsanders: These are just not defined on 0.4. Do you mean that I should change this to `VERSION <= v"0.4"`?

src/intervals/arithmetic.jl
dpsanders: You mean `isdefined(:Iterators)`? I don't know how to find the exact cutoff where this change was made. In any case, if it's a dev version isn't the expectation that people will keep updating it to the latest version?

src/intervals/arithmetic.jl
tkelman: No, I mean do you need to be extending a base method on base types? If these are backporting something that worked in later versions of Julia, is it available in Compat?

src/intervals/arithmetic.jl
tkelman: There should be a PR linked to in NEWS or otherwise not too hard to find, and there's the `contrib/commit-name.sh` script you can use

src/intervals/arithmetic.jl
dpsanders: These are indeed backporting something that is in 0.5 onwards. It is not available in Compat, but probably should be. https://github.com/JuliaLang/Compat.jl/issues/313

dpsanders: Updated in master to use the correct precise versions, thanks.

Issue #194, opened 2017-01-29
Hangs taking power of decorated interval


dpsanders:
```
julia> x = @decorated(-1, 2)
DecoratedInterval(Interval(-1.0, 2.0), com)

julia> x^Interval(3..4)
```
hangs

lbenet:
You are right; I'll try to take a look on this...

lbenet:
We had only implemented `x^y` with `x` and `y` being `DecoratedInterval`. See #195

dpsanders:
Fixed by #195

PR #195, opened 2017-01-30
Implement x^y, with x::DecoratedInterval and y::Interval

lbenet:
Solves #194


dpsanders:
Could you add a test, please?

lbenet:
I just added some tests; they are not too thorough, but the ITF1788 takes care about this
more systematically.

dpsanders:
Thanks!

Issue #193, opened 2017-01-28
Rewrite tests to use `Base.Test`


dpsanders:
This will involve redoing the ITF1788 output.

lbenet:
I don't disagree, but is there any reason for this?

dpsanders:
FactCheck is deprecated (now lives in JuliaArchive), and Base.Test now has the same functionality, and is more natural (`==` instead of `-->` and `begin` instead of `do` for `@testset`).

lbenet:
Thanks for the update. I surely agree.

PR #205, opened 2017-02-16
Rewrite tests to use Base.Test

lbenet:
Currently, some tests (involving some functions) do not pass due
to CRlibm

### DIFF
--- a/src/ValidatedNumerics.jl
+++ b/src/ValidatedNumerics.jl
@@ -37,7 +37,7 @@ export
     precedes, strictprecedes, ≺,
     entireinterval, isentire, nai, isnai, isthin, iscommon,
     widen, infimum, supremum,
-    parameters, eps, dist, roughly,
+    parameters, eps, dist, 
     pi_interval,
     midpoint_radius, interval_from_midpoint_radius,
     RoundTiesToEven, RoundTiesToAway,
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,2 +1,3 @@
 FactCheck 0.3
 Polynomials 0.0.5
+BaseTestNext
--- a/test/decoration_tests/decoration_tests.jl
+++ b/test/decoration_tests/decoration_tests.jl
@@ -1,4 +1,9 @@
-using Base.Test
+if VERSION >= v"0.5.0-dev+7720"
+    using Base.Test
+else
+    using BaseTestNext
+    const Test = BaseTestNext
+end
 using ValidatedNumerics
 
 @testset "DecoratedInterval tests" begin
--- a/test/decoration_tests/decoration_tests.jl
+++ b/test/decoration_tests/decoration_tests.jl
@@ -1,62 +1,50 @@
-
-using FactCheck
+using Base.Test
 using ValidatedNumerics
 
-facts("DecoratedInterval tests") do
+@testset "DecoratedInterval tests" begin
     a = DecoratedInterval(@interval(1, 2), com)
-    @fact decoration(a) --> com
+    @test decoration(a) == com
 
     b = sqrt(a)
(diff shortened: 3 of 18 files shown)


lbenet:
Addresses #193

ITF1788 files not yet modified, since this is better to using directly the tools provided by ITF1788...

lbenet:
Incidentally, these changes will only be useful once we leave 0.4 without support.

dpsanders:
Great, thanks! On Julia 0.4 you need `using BaseTestNext` to use `@testset`.

lbenet:
What about `@compat`?

dpsanders:
Not sure what you are asking about `@compat`?

dpsanders:
https://github.com/JuliaCI/BaseTestNext.jl

lbenet:
I though that `@compat using Base.Test` could save us replacing that for
```julia
if VERSION >= v"0.5.0-dev+7720"
    using Base.Test
else
    using BaseTestNext
    const Test = BaseTestNext
end
```
in each file. But i couldn't find it.

How do we deal with the `REQUIRES` in this case?

dpsanders:
I think you only need it in runtests.jl, since everything is included from
there?

lbenet:
No, though that could indeed be a solution; each included file has the `using` command.

dpsanders:
If I remember correctly, each file has it so that they can be run
independently if needed. But the tests should work on 0.4 just adding the
basetestnext in runtests.jl

dpsanders:
Looking good! Is this ready to merge?

lbenet:
Not yet; on the one hand, there is something weird with Julia 0.6; I'm also trying to transform the ITF1788 tests. I'll let you know when I think it is ready.

dpsanders:
Don't worry about 0.6. I finally got tests to pass on the fix_radius_0.6 branch that requires us to change from FixedSizeArrays.jl to StaticArrays.jl. 

I suggest we merge this and the itf1788 tests can be in a separate pr.

lbenet:
Ok. Then go ahead.

lbenet:
One of the tests that is not passing in v0.6 yields the following output
```
Operations on boxes: Error During Test
  Test threw an exception of type ErrorException
  Expression: 2A == IntervalBox(2 .. 4, 6 .. 8)
  type TypeName has no field primary
  Stacktrace:
   [1] promote_type_nested(::Type{T} where T<:Tuple) at /Users/benet/.julia/v0.6/FixedSizeArrays/src/core.jl:181
   [2] construct_similar(...) at /Users/benet/.julia/v0.6/FixedSizeArrays/src/core.jl:230
   [3] macro expansion at /Users/benet/.julia/v0.6/FixedSizeArrays/src/mapreduce.jl:81 [inlined]
   [4] map at /Users/benet/.julia/v0.6/FixedSizeArrays/src/mapreduce.jl:119 [inlined]
   [5] *(::Int64, ::ValidatedNumerics.IntervalBox{2,Float64}) at /Users/benet/.julia/v0.6/FixedSizeArrays/src/ops.jl:95
   [6] macro expansion at /Users/benet/.julia/v0.5/ValidatedNumerics/test/multidim_tests/multidim.jl:13 [inlined]
   [7] macro expansion at ./test.jl:853 [inlined]
   [8] anonymous at ./<missing>:?
```

I guess this is related to the new type nomenclature... 

I haven't checked if there are other problems.

dpsanders:
I'm not sure what exactly it's related to, but it's fixed by the latest StaticArrays.jl update.

dpsanders:
Thanks!

dpsanders:
Do you have the ITF1788 tests switched to `@testset` for the files that are already included (not the new ones)?

PR #207, opened 2017-02-18
WIP: Mechanism for changing global rounding mode

dpsanders:
To use run Julia as follows from the command line:

```
> VN_ROUNDING=fast julia --compilecache=no
```
The options are `fast`, `none` and `correct`.
The default is `correct`.

The `--compilecache=no` is required each time the mode is changed.
Currently the `VN_ROUNDING=...` is required even if the mode is unchanged.

Sample timings for the following benchmark function after warmup:
```julia
function bench(N)
    a = 0.1 .. 0.1
    sum(sqrt(n*a) for n in 1:N)
end

julia> bench(10^7)
```

- correct: 2.858399 seconds
- fast: 1.014718 seconds
- none:  0.878046 seconds

cc @lbenet

### DIFF
--- a/src/intervals/rounding.jl
+++ b/src/intervals/rounding.jl
@@ -2,75 +2,168 @@
 # E.g.  +(a, b, RoundDown)
 # Some, like sin(a, RoundDown)  are already defined in CRlibm
 
+# Rounding modes:
+# - :correct  # correct rounding
+# - :fast     # fast rounding by prevfloat and nextfloat  (slightly wider than needed)
+# - :none     # no rounding at all for speed
 
-# import Base: +, -, *, /, sin, sqrt, inv, ^, zero, convert, parse
-
-
-# unary minus:
--{T<:AbstractFloat}(a::T, ::RoundingMode) = -a  # ignore rounding
-
-# zero:
-zero{T<:AbstractFloat}(a::Interval{T}, ::RoundingMode) = zero(T)
-zero{T<:AbstractFloat}(::Type{T}, ::RoundingMode) = zero(T)
-
-convert(::Type{BigFloat}, x, rounding_mode::RoundingMode) = setrounding(BigFloat, rounding_mode) do
-    convert(BigFloat, x)
-end
-
-parse{T}(::Type{T}, x, rounding_mode::RoundingMode) = setrounding(T, rounding_mode) do
-    parse(T, x)
-end
-
-
-# no-ops for rational rounding:
-for f in (:+, :-, :*, :/)
-    @eval $f{T<:Rational}(a::T, b::T, ::RoundingMode) = $f(a, b)
-end
-
-sqrt{T<:Rational}(a::T, rounding_mode::RoundingMode) = setrounding(float(T), rounding_mode) do
-    sqrt(float(a))
-end
-
-
-
-for mode in (:Down, :Up)
-
-    mode1 = Expr(:quote, mode)
-    mode1 = :(::RoundingMode{$mode1})
-
-    mode2 = Symbol("Round", mode)
-
-
-    for f in (:+, :-, :*, :/,
-                :atan2)
-
-        @eval begin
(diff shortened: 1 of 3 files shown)


dpsanders:
Note, however, that apparently the elementary functions like `sin` are much faster in `CRlibm` than in base Julia!

src/intervals/rounding.jl
```diff
@@ -2,74 +2,176 @@
...
 
-# import Base: +, -, *, /, sin, sqrt, inv, ^, zero, convert, parse
 
+doc"""Choose rounding mode based on environment variable"""
 
-# unary minus:
--{T<:AbstractFloat}(a::T, ::RoundingMode) = -a  # ignore rounding
+function setup_rounding_mode()
+    if haskey(ENV, "VN_ROUNDING")
+        rounding_mode = Symbol(ENV["VN_ROUNDING"])
```
lbenet: Is this equivalent to the result of `get_rounding_mode()`?

lbenet: Another question: is this function *useful*? What I mean is that you have to start julia in a certain way (see https://github.com/dpsanders/ValidatedNumerics.jl/pull/207#issue-208675288), so my guess is that you cannot change interactively the rounding mode, or can you?

src/intervals/rounding.jl
```diff
@@ -2,74 +2,176 @@
...
 
-    @eval begin
-        function ^{T<:AbstractFloat,S}(a::T, b::S, $mode1)
-            setrounding(T, $mode2) do
-                ^(a, b)
-            end
-        end
+
+    # power:
+    @static if ROUNDING == :correct
```
lbenet: I guess that the use of `@static` is related to setting `VN_ROUNDING`?

dpsanders:
As far as I can tell, unfortunately the best / only way to be able to change the global rounding type is in this rather annoying way that cannot be done interactively.  The `@static if` checks things that are known at compile time. 

The idea is basically that the rounding mode change is implemented by defining functions like `sin(0.5, RoundDown)` differently, depending on the global rounding mode, which cannot be changed interactively, or at least not simply and not in 0.6. I am not sure that this is correct or the correct solution, but it seems to work, and allows us to check the effect of using different rounding scenarios.

src/intervals/rounding.jl
dpsanders: `@static` is for constants that are known at compile time.

src/intervals/rounding.jl
dpsanders: This function should be called only when the package is being loaded. The only (current) reason for making it into a function is for tidiness. I am hoping that at some point it will be possible to turn it into a function that may be used interactively, but according to the discussions I have had, it doesn't seem possible currently.

lbenet:
> I am not sure that this is correct or the correct solution, but it seems to work, and allows us to check the effect of using different rounding scenarios.

The problem I see here is that you are not only changing the behavior of `sin(::Interval)` but also `sin(::Number)`, which is kind of dangerous.

dpsanders:
No, I am only changing the behaviour of  `sin(x::Number, r::RoundingMode)`. These are functions that are currently only in `CRlibm` and here.

lbenet:
Good to hear that the change is only local and affects only CRlibm functions!

What happens if I start Julia as I do always, without the specifications of the environment variable `VN_ROUNDING` and `--compilecahe=no`? Do I get the `:correct` rounding mode, i.e., is that the version that is precompiled?

lbenet:
Is there a way of including tests with travis?

While I'm not against merging this (as you say, it seems to work), it would be nice to have tests it more thoroughly.

dpsanders:
If you start it as you usually do, then you should get the `:correct` rounding mode, yes.

I do not have any idea how to test this on Travis. I will try to find out.

dpsanders:
I think I now have a better way of implementing this, so closing.

PR #208, opened 2017-02-18
Add fast integer power function

dpsanders:
Add a fast integer power function, `pow`.

### DIFF
--- a/src/intervals/functions.jl
+++ b/src/intervals/functions.jl
@@ -167,7 +167,11 @@ function sqrt{T}(a::Interval{T})
     @round(sqrt(a.lo), sqrt(a.hi))  # `sqrt` is correctly-rounded
 end
 
+doc"""
+    pow(x::Interval, n::Integer)
 
+A faster implementation of `x^n` using `power_by_squaring`.
+"""    
 function pow{T}(x::Interval{T}, n::Integer)  # fast integer power
 
     isempty(x) && return x
--- a/src/intervals/functions.jl
+++ b/src/intervals/functions.jl
@@ -168,6 +168,29 @@ function sqrt{T}(a::Interval{T})
 end
 
 
+function pow{T}(x::Interval{T}, n::Integer)  # fast integer power
+
+    isempty(x) && return x
+
+    if iseven(n) && zero(T) ∈ x
+
+        return hull(zero(x),
+                    hull(Base.power_by_squaring(Interval(mig(x)), n),
+                        Base.power_by_squaring(Interval(mag(x)), n))
+            )
+
+    else
+
+      return hull( Base.power_by_squaring(Interval(x.lo), n),
+                    Base.power_by_squaring(Interval(x.hi), n) )
+
+    end
+
+end
+
+
+
+
 for f in (:exp, :expm1)
     @eval begin
         function ($f){T}(a::Interval{T})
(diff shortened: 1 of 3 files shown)


src/intervals/functions.jl
```diff
@@ -168,6 +168,29 @@ function sqrt{T}(a::Interval{T})
 end
 
 
+function pow{T}(x::Interval{T}, n::Integer)  # fast integer power
```
lbenet: Please, add docstrings!

test/interval_tests/numeric.jl
```diff
@@ -197,3 +197,17 @@ end
...
+    x = 1..2
+    @test pow(x, 2) == pow(-x, 2) == Interval(1, 4)
+    @test pow(-x, 3) == Interval(-8.0, -1.0)
+
+    @test pow(-1..2, 2) == 0..4
+    @test pow(-1..2, 3) == -1..8
+    @test pow(-1..2, 4) == 0..16
+
+    x = @interval(pi)
+    @test x^100 ⊆ pow(x, 100)
```
lbenet: I do like this test! Yet, I think this should be extende to include tests which play a little bit with the rounding or with other `Interval{T}` types. 

It would also be nice to have an example where `interior(a^n, pow(a,n))` is `true`.

dpsanders:
I've added a docstring and a few tests. Not sure what you mean about the rounding mode.

Some benchmarks using `BenchmarkTools.jl`:

```

julia v0.5> @btime x^2
  2.343 μs (44 allocations: 1.73 KiB)
[1, 4]

julia v0.5> @btime pow(x, 2)
  385.174 ns (9 allocations: 160 bytes)
[1, 4]

julia v0.5> @btime x^10
  3.050 μs (46 allocations: 1.87 KiB)
[1, 1024]

julia v0.5> @btime pow(x, 10)
  887.837 ns (21 allocations: 352 bytes)
[1, 1024]

julia v0.5> @btime pow(x, 10)^C

julia v0.5> @btime x^100
  3.609 μs (46 allocations: 1.87 KiB)
[1, 1.26766e+30]

julia v0.5> @btime pow(x, 100)
  1.554 μs (37 allocations: 608 bytes)
[1, 1.26766e+30]
```
As expected, it performs much better for small powers, and only a bit better (but still > twice as fast) for large powers.

dpsanders:
By the way, the function `interior` has no docstring ;) 
And the name of the function does not indicate that it returns a boolean. Maybe change to `isinterior` to follow the Julia convention for functions returning bools? (And any other similar functions.)

src/intervals/functions.jl
```diff
@@ -167,6 +167,33 @@ function sqrt{T}(a::Interval{T})
     @round(sqrt(a.lo), sqrt(a.hi))  # `sqrt` is correctly-rounded
 end
 
+doc"""
+    pow(x::Interval, n::Integer)
+
+A faster implementation of `x^n` using `power_by_squaring`.
```
lbenet: Thanks! I think it should also be stated that it *may* yield broader intervals than `x^n`.

lbenet:
> By the way, the function `interior` has no docstring ;)
And the name of the function does not indicate that it returns a boolean. Maybe change to `isinterior` to follow the Julia convention for functions returning bools? (And any other similar functions.)

You right about the docstrings, but maybe this should be the contents of another PR. I also like your suggestion about `isinterior` instead of `interior`, which should also be applied to other such functions. Originally, I used the same name used in the standard, but we may change the name.

lbenet:
I just noticed #219 ...

lbenet:
Great work! I'm merging!

lbenet:
(Sorry, wrong button.)

lbenet:
Now it's merged...

Issue #172, opened 2016-11-05
Using ± in construction of intervals ignores @bigintervals


gmoutsin:
Writing

`@biginterval(1)±1`

returns

`[-0, 2]`

Which is of type `ValidatedNumerics.Interval{Float64}` instead of `ValidatedNumerics.Interval{BigFloat}` that is should be.

gmoutsin:
Setting the precision with `setprecision(Interval, 256)` works though.

dpsanders:
Thanks, this is a problem with the definition of `..`, which is currently used in `±`.
This, in turn, calls `@interval(a-b, a+b)`, which has the same problem, since `@interval` checks the global precision mode to see which kind of `Interval` to make. 

Solving the problem with `@interval` may be difficult; solving `..` and `±` should be easy.

lbenet:
Closing this; solved by #209

dpsanders:
Looking at this again, I don't think it makes very much sense to use `±` with interval arguments.

gmoutsin:
I did find it convenient to use as I had the value and the numerical error separately.

Issue #203, opened 2017-02-13
Implement `convert(Interval, x)`


dpsanders:
The correct type can be chosen as follows:
```
julia> Base.convert(::Type{Interval}, x::Real) = (T = typeof(float(x)); convert(Interval{T}, x))
```
which gives
```
julia> convert(Interval, big(3.))
[3, 3]₂₅₆

julia> convert(Interval, 3.1)
Interval(3.0999999999999996, 3.1)

julia> convert(Interval, 3.)
Interval(3.0, 3.0)
```

This can then be used to define `a..b` as
```
julia> a..b = hull(convert(Interval, a), convert(Interval, b))
```
which fixes #172.

lbenet:
Closing, since this is solved by #209.

PR #209, opened 2017-02-18
Rewrite .. operator

dpsanders:
Rewrites `a..b` to not use `@interval`.
Does this by using `convert(Interval, a)` and `convert(Interval, b)`.


dpsanders:
Fixes #172 and #203.

lbenet:
Merged! Thanks!

Issue #175, opened 2016-11-06
displaymode(format=:midpoint) does not work with biginterval


gmoutsin:
After `displaymode(format=:midpoint)` the command `@interval(1,2)`  returns `1.5 ± 0.5` but `@biginterval(1,2)` returns `nothing`.

dpsanders:
Fixed by #210.

Issue #131, opened 2016-05-13
showall not working correctly for BigFloat


dpsanders:
``` julia
julia> a = DecoratedInterval(big(2), big(3), com)
[2, 3]₂₅₆

julia> showall(a)  # wrong
[2, 3]₂₅₆

julia> displaymode(format=:midpoint)  # wrong

julia> a
nothing
```

`format=:full` is "correct":

``` julia
julia> a
DecoratedInterval(Interval(2.000000000000000000000000000000000000000000000000000000000000000000000000000000, 3.000000000000000000000000000000000000000000000000000000000000000000000000000000), com)
```

However, it does not have the desired effect of being able to copy and paste this into the REPL. 
This would require the output to have `big"2.000000000000000000000000000000000000000000000000000000000000000000000000000000"`

dpsanders:
This is mostly fixed in #146. 
Currently get 
```DecoratedInterval(Interval(2.000000000000000000000000000000000000000000000000000000000000000000000000000000, 3.000000000000000000000000000000000000000000000000000000000000000000000000000000)₂₅₆, com)
```
which is not copyable-and-pastable.

dpsanders:
Fixed by #210.

PR #210, opened 2017-02-18
Rewrite display.jl

### LINKED ISSUE #202: Rewrite display.jl
To fix various output issues and avoid using `invoke`.

dpsanders:
Fixes #202, #175, #131.

This changes the name of `displaymode` to `setdisplay`, and the syntax to the simpler

```
setdisplay(:full)
```
instead of the wordy
```
displaymode(format=:full)
```

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -6,7 +6,6 @@ end
 
 const display_params = DisplayParameters(:standard, false, 6)
 
-const display_options = [:standard, :full, :midpoint]
 
 doc"""
     setdisplay(;kw)
@@ -20,7 +19,7 @@ The following options are available:
     - `:full`: `Interval(1, 2)`
     - `:midpoint`: 1.5 ± 0.5
 
-- `sigfigs`: number of significant figures to show in `standard` mode
+- `sigfigs`: number of significant figures to show in `standard` mode; the default is 6
 
 - `decorations` (boolean):  show decorations or not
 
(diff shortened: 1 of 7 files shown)


src/display.jl
```diff
@@ -23,8 +23,13 @@ The following options are available:
 - `sigfigs`: number of significant figures to show in `standard` mode
```
lbenet: I think it is worth writing here that the default number is 6.

src/display.jl
```diff
@@ -23,8 +23,13 @@ The following options are available:
...
 
 - `decorations` (boolean):  show decorations or not
+
+Example:
+```
+julia> setdisplay(:full, decorations=true)
+```
 """
-function displaymode(;decorations=nothing, format=nothing, sigfigs=-1)
+function setdisplay(format=nothing; decorations=nothing, sigfigs::Integer=-1)
```
lbenet: I guess there is a reason why the default is `nothing` for each kw, but why is this better than setting them directly to the values specified in `display_params`?

src/display.jl
```diff
@@ -36,7 +41,13 @@ function displaymode(;decorations=nothing, format=nothing, sigfigs=-1)
...
+
+        if decorations ∈ (true, false)
+            display_params.decorations = decorations
+
+        else
+            throw(ArgumentError("`decorations` must be `true` or `false`"))
+        end
     end
 
     if sigfigs >= 0
```
lbenet: What is the meaning of `sigfigs=0`? Anywhay, I get the following:
```julia
julia> @interval(0.1, 0.2)
[0.0999999, 0.200001]

julia> setdisplay(:standard, sigfigs=0)
0

julia> @interval(0.1, 0.2)
[0.09, 0.3]
```

src/display.jl
```diff
@@ -133,6 +146,9 @@ function subscriptify(n::Int)
 end
 
 
+# fall-back:
```
lbenet: Why do you need `basic_representation` *and* `representation`, or isn't it enough with one function and several methods?

src/display.jl
dpsanders: Good point, done.

src/display.jl
dpsanders: That's a good idea, thanks.

src/display.jl
dpsanders: I have made 1 the minimum value of sigfigs.

src/display.jl
dpsanders: I had just `representation` before. The problem is that `representation(x::Interval{BigFloat})` calls `representation` for the interval part and then adds the part with the number of digits. 

To do so, I had to use `invoke`, which is ugly. The simple solution was to add `basic_representation` instead.

dpsanders:
I have made the changes you suggested, thanks. I'll merge once it goes green.

src/display.jl
lbenet: Thanks for the explanation.

PR #211, opened 2017-02-19
Convert x in Interval{T}(x) before passing to Interval

### LINKED ISSUE #170: Eliminate extra `convert` when creating an interval
``` julia
julia> Interval{BigFloat}(0.1)
[0.0999999, 0.100001]₂₅₆

julia> showall(ans)
Interval(9.999999999999999999999999999999999999999999999999999999999999999999999999999914e-02, 1.000000000000000000000000000000000000000000000000000000000000000000000000000002e-01)
```

This is doing `convert(Interval{BigFloat}, 0.1)`.
Probably it should instead just do the equivalent of `Interval{BigFloat}(BigFloat(0.1))`.

This can be achieved by adding the following:

``` julia
(::Type{Interval{T}}){T}(arg) = Interval(convert(T, arg))
```

where the `convert` to type `T` is done directly on the argument.

dpsanders:
Fixes #170.


lbenet:
Merged! Thanks.

PR #212, opened 2017-02-19
Do not allow 2 NaNs in Interval

### LINKED ISSUE #192: Do not allow `Interval(NaN, 2)`
If one of the entries is `NaN`, probably the other should be too.

dpsanders:
Fixes #192.

Note that this contains also the changes in `rewrite_display` and should be merged after that.

### DIFF
--- a/test/interval_tests/construction.jl
+++ b/test/interval_tests/construction.jl
@@ -248,6 +248,9 @@ end
 
 # issue 192:
 @testset "Disallow a single NaN in an interval" begin
-    @test Interval(NaN, 2) == Interval(NaN, NaN)
-    @test Interval(Inf, NaN) == Interval(NaN, NaN)
+    a = Interval(NaN, 2)
+    @test isnai(a)
+
+    a = Interval(Inf, NaN)
+    @test isnai(a)
 end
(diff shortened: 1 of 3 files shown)


lbenet:
This branch is conflicting with current master, specifically in `src/display.jl`.

test/interval_tests/construction.jl
```diff
@@ -196,3 +196,12 @@ end
     @test 3 ± 0.1 == Interval(2.9, 3.1)
     @test 0.5 ± 1 == Interval(-0.5, 1.5)
 end
+
+# issue 192:
+@testset "Disallow a single NaN in an interval" begin
+    a = Interval(NaN, 2)
+    @test isnai(a)
```
lbenet: Since `isnai(a)` verifies that either `a.lo` or `a.hi` is `NaN`, this tests does not quite guarantee that *both* are `NaN`s. I suggest adding another test of the form `@test isnan(a.lo) && isnan(a.hi)`, which precisely addresses this. Once said this, I don't think we need to change `isnai()`...

dpsanders: Good point, thanks.

lbenet:
Thanks! I'm merging!

PR #216, opened 2017-02-19
Add iseven and isodd for intervals

dpsanders:


lbenet:
Can you explain what is the motivation to define this?

dpsanders:
It is used in `IntervalConstraintProgramming.jl` to calculate the correct inverse of `x^n` for integer `n`. Thinking about it, that is probably not a very efficient approach (converting an integer to an interval and then checking if it's even), but it is the simplest solution there.

In any case, it seems useful to define these functions here.

lbenet:
Why not treating it simply as an `Int`? 

I imagine that you eventually need to make it interval because a power `1/n` appears for the inverse function; am I right?

dpsanders:
I should probably just treat it as an `Int` yes, but for code simplicity it is currently treated as an interval.

dpsanders:
I think you are right that this is not necessary; closing.

PR #220, opened 2017-02-22
Rounding modes using traits

dpsanders:
This is an implementation of changing the global rounding mode using "traits".

Note that **actually changing the rounding mode for calculations works *only* on Julia v0.6**, since it depends on the solution to https://github.com/JuliaLang/julia/issues/265.

From the point of view of the user, it reduces to doing

```
setrounding(Interval, :correct)
```

I have not yet added tests.

### DIFF
--- a/src/ValidatedNumerics.jl
+++ b/src/ValidatedNumerics.jl
@@ -4,7 +4,7 @@ __precompile__(true)
 
 module ValidatedNumerics
 
-using CRlibm
+import CRlibm
 using Compat
 using StaticArrays
 using ForwardDiff
--- a/src/intervals/precision.jl
+++ b/src/intervals/precision.jl
@@ -22,6 +22,14 @@ function big53(a::Interval{Float64})
     end
 end
 
+function big53(x::Float64)
+    # BigFloat(x, 53)  # in Julia v0.6
+
+    setprecision(53) do
+        BigFloat(x)
+    end
+end
+
 
 setprecision(::Type{Interval}, ::Type{Float64}) = parameters.precision_type = Float64
 # does not change the BigFloat precision
--- a/src/intervals/precision.jl
+++ b/src/intervals/precision.jl
@@ -85,44 +85,13 @@ Base.float{T}(::Type{Rational{T}}) = typeof(float(one(Rational{T})))
 
 # Use that type for rounding with rationals, e.g. for sqrt:
 
-if VERSION < v"0.5.0-dev+1182"
-
-    function Base.with_rounding{T}(f::Function, ::Type{Rational{T}},
-        rounding_mode::RoundingMode)
-        setrounding(f, float(Rational{T}), rounding_mode)
-    end
-
-else
-    function Base.setrounding{T}(f::Function, ::Type{Rational{T}},
-        rounding_mode::RoundingMode)
-        setrounding(f, float(Rational{T}), rounding_mode)
-    end
+function Base.setrounding{T}(f::Function, ::Type{Rational{T}},
+    rounding_mode::RoundingMode)
+    setrounding(f, float(Rational{T}), rounding_mode)
 end
+
 
 
 float{T}(x::Interval{T}) = convert(Interval{float(T)}, x)  # https://github.com/dpsanders/ValidatedNumerics.jl/issues/174
-
-## Change type of interval rounding:
-
-
-doc"""`rounding(Interval)` returns the current interval rounding mode.
-There are two possible rounding modes:
-
-- :narrow  -- changes the floating-point rounding mode to `RoundUp` and `RoundDown`.
-This gives the narrowest possible interval.
-
-- :wide -- Leaves the floating-point rounding mode in `RoundNearest` and uses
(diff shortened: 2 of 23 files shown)


src/ValidatedNumerics.jl
```diff
@@ -4,7 +4,7 @@ __precompile__(true)
 
 module ValidatedNumerics
 
-using CRlibm
+import CRlibm
```
lbenet: There is a bunch of warnings appearing because methods (of CRlibm) are overwritten in `src/intervals/rounding.jl`. I guess this is partially due to this `import`, or is it that some functions are truly overwritten?

src/intervals/rounding.jl
```diff
@@ -2,74 +2,191 @@
...
-# zero:
-zero{T<:AbstractFloat}(a::Interval{T}, ::RoundingMode) = zero(T)
-zero{T<:AbstractFloat}(::Type{T}, ::RoundingMode) = zero(T)
 
-convert(::Type{BigFloat}, x, rounding_mode::RoundingMode) = setrounding(BigFloat, rounding_mode) do
-    convert(BigFloat, x)
-end
+doc"""Choose rounding mode based on environment variable"""
+
+immutable RoundingType{T} end
```
lbenet: I guess there is no need to impose that `T<:Symbol`... Or is it worth imposing it?

src/intervals/rounding.jl
```diff
@@ -2,74 +2,191 @@
...
+    end
+
+    if rounding_type ∉ (:correct, :fast, :none)
+        throw(ArgumentError("Rounding type must be one of `:correct`, `:fast`, `:none`"))
+    end
+
+    #rounding_type = :(RoundingType{$(Meta.quot(rounding_type))})
+    rounding_object = RoundingType{rounding_type}()
+
+
```
lbenet: Do we need to overwrite the functions below? I think some of them are defined in the loop above, aren't they?

src/intervals/rounding.jl
dpsanders: Not really a need, but maybe clearer -- done, thanks.

dpsanders:
I changed `RoundingType` --> `IntervalRounding` for clarity and precision

src/intervals/rounding.jl
dpsanders: The design is as follows: All functions are defined like `sin(::IntervalRounding{:correct}, x, ::RoundingMode)` in the main body of the file. Functions with different interval rounding types *coexist*, and are always all present.

In the function `setrounding(Interval, rounding_type)`, the functions like `sin(x, ::RoundingMode)` are defined, i.e. *without* the explicit interval rounding type. These functions get redefined when you call `setrounding(Interval, rounding_type)` again.

In Julia v0.6 (but *not* in 0.5), this will *automatically redefine all relevant functions*, in particular those used in `+(a::Interval, b::Interval)` etc., so that all interval functions will automatically work with the correct interval rounding type!

I will add the above explanation at the top of the file.

dpsanders:
I think the CRlibm problem was because it requires CRlibm v0.5. This should now be fixed.

dpsanders:
This branch is aimed only at post-Julia-0.4.

dpsanders:
This is ready. All tests pass on 0.5, and all tests except multidimensional pass on 0.6; these are waiting for StaticArrays improvements.

I added a basic test in interval_tests/rounding.jl  for changing the rounding mode.
This checks that the sin function is correctly redefined on 0.6 (and is only run on 0.6) and tests pass there.  The tests are tricky to get right due to "world-age" problems (the fix of Julia issue #265 on redefining dependent functions.)

I would like to merge and release 0.8.

lbenet:
Merging...

src/intervals/precision.jl
```diff
@@ -77,44 +85,13 @@ Base.float{T}(::Type{Rational{T}}) = typeof(float(one(Rational{T})))
...
-        rounding_mode::RoundingMode)
-        setrounding(f, float(Rational{T}), rounding_mode)
-    end
-
-else
-    function Base.setrounding{T}(f::Function, ::Type{Rational{T}},
-        rounding_mode::RoundingMode)
-        setrounding(f, float(Rational{T}), rounding_mode)
-    end
+function Base.setrounding{T}(f::Function, ::Type{Rational{T}},
```
tkelman: what's the reason for this? should `T` be constrained to a package type, or this be modified in base?

dpsanders: It's just a way to say that rounding is a no-op for rationals.

Currently in Julia `setrounding(T, RoundDown)` is an error for `T` that are not `<: AbstractFloat`, e.g.

```
julia> setrounding(Int, RoundDown)
```

Since we have `Interval`s of `Rational`s, we need this no-op method for code reusability.

dpsanders: I'm not sure to what extent it's useful to have this in Base. We're probably the only people that "need" to have this be a no-op. But maybe it's worth defining a no-op fallback.

tkelman: it's at least worth opening a discussion of what base should do, the behaviors there (and use cases for them) might not all be deeply thought out

PR #222, opened 2017-02-25
Add NEWS for 0.7 release

dpsanders:


lbenet:
Either on v0.6 or v0.7, a comment should be added about which is the version that is the last to support Julia 0.4.

dpsanders:
Thanks for the comment. 0.7 will be the last release to support Julia v0.4; I have added this to the NEWS file.

PR #217, opened 2017-02-19
Add bisection of Intervals and IntervalBoxes

dpsanders:

PR #223, opened 2017-02-25
Revert "Add bisection of Intervals and IntervalBoxes"

dpsanders:
Reverts dpsanders/ValidatedNumerics.jl#217

This worked fine on the branch, but gives a very strange error and failing tests on master. 
I don't understand why.

I suggest reverting this for now and trying again with `StaticArrays`.


dpsanders:
Merging since tests are failing on master. cc @lbenet

PR #224, opened 2017-02-25
Deprecate displaymode

dpsanders:


lbenet:
LGTM. This should also be merged.

PR #225, opened 2017-02-25
Redo bisect

dpsanders:
Redoing #217 that was reverted in #223


dpsanders:
Do not merge for the moment.

lbenet:
Merging #227 I think this is automatically taken into account, am I right?

dpsanders:
I included #227 in here to make sure that it did fix the problem.
I had reverted the bisection PR, so this PR will need to be merged once #227 is.

PR #227, opened 2017-03-04
Fix convert(Interval{Float64}, x::Interval)

dpsanders:


src/intervals/intervals.jl
```diff
@@ -47,7 +47,9 @@ eltype{T<:Real}(x::Interval{T}) = T
 Interval(x::Interval) = x
 Interval(x::Complex) = Interval(real(x)) + im*Interval(imag(x))
 
-@compat (::Type{Interval{T}}){T}(arg) = Interval(convert(T, arg))
+@compat (::Type{Interval{T}}){T}(x) = Interval(convert(T, x))
+
+@compat (::Type{Interval{T}}){T}(x::Interval) = convert(Interval{T}, x)
```
lbenet: Would it be convenient to have some sort of diagonal dispatch here, I mean:
```julia
@compat (::Type{Interval{T}}){T}(x::Interval) = convert(Interval{T}, x)
@compat (::Type{Interval{T}}){T}(x::Interval{T}) = x
```

dpsanders: Very good question. It seems that
`convert(Interval{T}, x::Interval{T})` is not currently defined. I'll add it.

PR #229, opened 2017-03-04
Add test of at-interval with fields from a type

dpsanders:


dpsanders:
Rebased and should be ready once goes green.

PR #235, opened 2017-03-05
Adapt IT1788 tests to use Base.Test and run them in parallel

lbenet:
Note that in ITF1788_tests.jl does not yet include running the tests in
ieee1788-constructors.jl; some of these tests do not pass yet, since
some functionality is lacking.


lbenet:
This completes #205

lbenet:
Checking what went wrong...

dpsanders:
This is great work! 

What does addprocs() with no arguments do?

dpsanders:
The failing tests seem to be since a file is missing.

lbenet:
Thanks!

It adds all possible workers of your machine (worker 1 is master), so you use them all to do the tests.

lbenet:
Some tests, not (yet) related to the ITF1788 suite are not passing in travis.

lbenet:
Sorry, it is indeed related to the ITF1788 suite....

dpsanders:
It looks like you forgot to commit the bool test file?

lbenet:
But I see it there. Maybe a question of path in travis...?

dpsanders:
Not sure.

lbenet:
Indeed, it is the path: It is looking for "/home/travis/.julia/v0.5/ValidatedNumerics/test/libieeep1788_tests_bool.jl" while it should be "/home/travis/.julia/v0.5/ValidatedNumerics/test/ITF1788_tests/libieeep1788_tests_bool.jl"

dpsanders:
Ah OK, so you can just add the directory name. Or. Maybe do a "cd". 

It is maybe  worth looking in the Julia base tests to see how they do it to be compatible with Windows at some point in the future. Or just leave it unixy for now for simplicity.

dpsanders:
Ah you can just use joinpath

lbenet:
I just pushed a new commit that should solve this; if you tell me how to use it, I do change it

lbenet:
Another try...

lbenet:
Tests are running.... finally!

dpsanders:
Great! Is it possible to parallelise the non itf tests (ours) too?

lbenet:
The scheme we have now is pretty naive: each worker includes a test file independently. 

This can be extended for all included test files, but we must be sure about two things: the master (worker 1) has to be in charge of the very first test, since it has to precompile. The second thing is to have all workers do the ITF1788 tests, which take really long.

Shall I implement this in this PR?

dpsanders:
I suggest you make a new pr in a new branch based in this branch. In that way we can easily see. If it has any effect on the total time to run the tests

lbenet:
Travis is giving problems: while tests pass in 0.4, they take too long in 0.5; in 0.6, I think the problem is different...

lbenet:
Maybe we shouldn't wait for this to be ready for 0.7...

lbenet:
Locally, it seems something happens with the "minimal_pow_test"... they are many, but are taking really too long.

dpsanders:
0.6 will require StaticArrays.jl after we release 0.7

I don't think the problem on 0.5 was that it took too long. It seems not to have realised that the tests finished.

dpsanders:
Yes that one was always problematical. One solution is to split it into
several files. The problem is that that is not so easy to do automatically.
We could write a script to do it.

lbenet:
No. travis says that it doesn't receive any output for more than 10 min. Yet, the next item ("minimal_fma_test") is not displayed, so it was in process. It corresponds to 564 tests...

lbenet:
Thanks! See also #236, with similar problems...

dpsanders:
The problem as usual seems to be the huge size of the `pow` tests.
But I don't understand why they were working and are now not working again.

lbenet:
Locally, they work. But they take an *enormous* amount of time to complete. May this be related to Base.Test?

lbenet:
We may still use FactCheck.jl, but implement that in parallel...

lbenet:
Why not breaking those tests in few files?

dpsanders:
Yes, breaking up the files would be the best solution. But it seems difficult to automate with ITF1788?

lbenet:
I agree...

lbenet:
I'm not sure what are you aiming...

dpsanders:
I'm trying to make travis wait longer and not kill the tests when there's
no output. It's not working yet though.

lbenet:
What ever you did seems to have worked! You should push another commit to get the correct code-coverage
.

dpsanders:
No, it's not actually running the tests...

dpsanders:
Given the unbelievably long time that the ITF1788 tests require with `Base.Test`on Julia 0.5 (hopefully 0.6 will be faster), maybe we should just leave the FactCheck versions for now, and parallelize those?

I was pointed to `Distributions.jl` as having a simple version of parallelized tests. It is a bit different from what you did:

https://github.com/JuliaStats/Distributions.jl/blob/master/test/runtests.jl

lbenet:
I've just checked the way they parallelize and the sole difference is that they use `pmap` (parallel map) instead of `@sync @parallel for`. I'm not quite sure, but I think it does exactly the same.

dpsanders:
What is the status of this?

lbenet:
I am not quite sure, but I think this is a leftover when `ValidetedNumerics.jl` became an umbrella package. I think we should close this and, in case of need reopen it in `IntervalArithmetics.jl`, where this belongs.

PR #245, opened 2017-03-06
Change from FixedSizeArrays.jl to StaticArrays.jl

dpsanders:


dpsanders:
Tests pass on Julia 0.5. 
0.6 requires changes in StaticArrays: https://github.com/JuliaArrays/StaticArrays.jl/issues/113

src/ValidatedNumerics.jl
```diff
@@ -26,8 +26,7 @@ import Base:
     precision,
     isfinite, isnan,
     show, showall,
-    isinteger, setdiff,
-    parse
```
lbenet: Don't we need to import `parse`? I though that it was extended...

dpsanders: Good catch, thanks! I think this bug was introduced in some rebase operation.

lbenet:
LGTM. Should I merge this?

dpsanders:
Can you add back in the import of parse? Does this not cause an error?

lbenet:
Sure. Also, I think we also want to import `setdiff`... not sure about `isinteger` though.

lbenet:
Just pushed a commit reimporting `parse`; the others were there...

dpsanders:
Yes to setdiff and isinteger, they are both defined for Intervals if I remember correctly (currently away from computer).

lbenet:
> Yes to setdiff and isinteger, they are both defined for Intervals if I remember correctly (currently away from computer).

I thought so...

lbenet:
Tests seem to pass, though there are "Allowed failures for 0.6"... Locally, there are failures related to `IntervalBox`. This is what I get:
```julia
Operations on boxes: Error During Test
  Test threw an exception of type ErrorException
  Expression: 2A == IntervalBox(2 .. 4, 6 .. 8)
  The size of type `ValidatedNumerics.IntervalBox{2,Float64}` is not known.
  
  If you were trying to construct (or `convert` to) a `StaticArray` you
  may need to add the size explicitly as a type parameter so its size is
  inferrable to the Julia compiler (or performance would be terrible). For
  example, you might try
  
      m = zeros(3,3)
      SMatrix(m)      # this error
      SMatrix{3,3}(m) # correct - size is inferrable
```

So I won't do anything here.

PR #247, opened 2017-03-07
Break up long testsets

cpierard:
Break chunks of code that are longer than 100 lines in the file `libieeep1788_tests_elem.jl`.


dpsanders:
Thanks! Did you do this by hand or in an automated way?

cpierard:
I did it by hand. I tried doing it in an automated way, but I failed.

dpsanders:
These files are automatically generated, so the next time they are regenerated all of your work will be lost. That's why this should be done in an automated way. 

What did you try?

cpierard:
Didn't know that. This is what I did so far:
```julia
f = open("libieeep1788_tests_elem.jl")
code_lines = readlines(f)

#println(code_lines[70][1:5])
#println(length(code_lines))

A = Int[]

chunk_begin = 0
chunk_end = 0

for i in 1:length(code_lines)

  if length(code_lines[i]) <= 3

    nothing

  elseif code_lines[i][1:3] == "fac"

    chunk_begin = i
    #println(chunk_begin)

  elseif code_lines[i][1:3] == "end"

    chunk_end = i
    #println(chunk_end)

  end

  if chunk_end - chunk_begin > 100

    push!(A, chunk_begin)
    chunk_end = 0

  end

end

println(A)
```
This code looks for the big blocks of code bigger than 100. My idea was to break-up the blocks in smaller ones by inserting `facts(... ) do` and `end` between the big blocks in the array that contains all the lines of code.

Do you have any suggestions?

dpsanders:
That looks like a good start.
You can use the `startswith` function to look at the start of a line.

I suggest you just accumulate the lines in each `facts` block into an array, and then write them out in blocks of 100 in new `facts` blocks.

dpsanders:
(Of course, 100 should be a variable that can be changed.)

cpierard:
Ok, I will wrap it in a function once is working well. What did you mean by expressions? Can you read an entire file as an expression (or an array of expressions)?

dpsanders:
I meant "regular expression" -- it's a way to specify a pattern to match and to capture the result of matching that pattern:

https://en.wikipedia.org/wiki/Regular_expression

https://en.wikibooks.org/wiki/Introducing_Julia/Strings_and_characters#Regular_expressions

cpierard:
thanks! It's just what I was looking for this morning. It will make the code look cleaner.

lbenet:
Would it be possible to get rid of a bunch of empty lines in `test/ITF1788_tests/test_file.jl `?

cpierard:
@lbenet Yes, it was a problem with the delimiter in the instruction `join(code_lines_new, "\t")`. Now is writing the file without empty lines.

cpierard:
I think the function is finished . I was comparing the `test_file.jl` with the original file, and it seems that all the lines of code are in the test file.  Maybe the name for the function is not the best...

dpsanders:
Thanks, it's looking good!
So in the end you didn't use regular expressions?

Maybe a better name is something like `split_testsets` (for the function and the filename).

Could you write a test for this? I suggest using a simple test file like the one here: https://github.com/JuliaLang/julia/issues/15346

The next step would be to make a new file that runs through each of the test files in the directory and splits them with a sensible name, maybe as they are being run.

I wonder if it would even be sensible to make this into a separate package, though that can be done later.

cpierard:
I didn't include my attempts to use regex because they were not working. So far I found the expression to match the title for each block:

`match(r"\".*\"", code_lines[chunk_begin]).match`

But returns `"\"minimal_pos_test\""`, and I have to get rid of the extras `"`.

I found two solutions:

`title = parse(match(r"\".*\"", code_lines[chunk_begin]).match)`

or

`title = split(match(r"\".*\"", code_lines[chunk_begin]).match, "\"")[2]`.

Ok, I will start writing the test for this.

dpsanders:
In the latest versions of Julia, the tests run much faster, so this is probably no longer necessary.

PR #254, opened 2017-03-14
Fix 1D Newton by checking correctly whether midpoint is exactly a root

dpsanders:
- Fixes Newton by correctly checking if a midpoint is exactly a root, and moving the midpoint a bit if it is.

- Fixes Krawczyk in the case that the midpoint has derivative 0.


lbenet:
Tests are not passing; more precisely, the root-finding tests.

dpsanders:
Tests are now passing! @lbenet

dpsanders:
This is ready. Is it OK with you to merge @lbenet ?

lbenet:
I'm also in favor of merging. Yet, some tests (0.6) are problematic, which is known. Do we go ahead?

dpsanders:
Yes, 0.6 is simply not ready yet. (it is still alpha after all)

lbenet:
Thanks! This was truly needed!

dpsanders:
Thanks!

PR #256, opened 2017-03-16
Added NEWS for 0.8

dpsanders:


lbenet:
I think this should be merged just before 0.8 is tagged.

lbenet:
Note that there are conflicts now with `NEWS.md`...

PR #258, opened 2017-03-17
0.6 update: type -> mutable struct

miguelraz:
I believe this should get this script up to speed with [version 0.6](https://github.com/JuliaLang/julia/blob/master/NEWS.md)

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -1,4 +1,4 @@
-type DisplayParameters
+mutable struct DisplayParameters
     format::Symbol
     decorations::Bool
     sigfigs::Int


dpsanders:
Thanks, but what about `immutable`?
There are also several warnings (at the end of the long list of warnings coming from `StaticArrays` about the new constructor syntax.

dpsanders:
And as @lbenet has commented, it's not clear what to do about Julia v0.5.
Is there information about that somewhere?

miguelraz:
[Here they are].(https://github.com/JuliaLang/julia/blob/master/NEWS.md)

ChrisRackauckas:
You shouldn't do this until you are ready to drop v0.5. Instead, use `@compat type ... end` with Compat 0.18.0 or higher.

dpsanders:
Thanks for the info, @ChrisRackauckas! We won't be dropping Julia v0.5 for a good while, I would think.

dpsanders:
In fact, `type` is still valid syntax on 0.6, no?

ChrisRackauckas:
>In fact, type is still valid syntax on 0.6, no?

Yes, and it still creates blocks. This is because `abstract type ... end` exists. So it parses correctly, which is why `@compat type ... end` is able to work.

lbenet:
I agree with David; we shouldn't drop support for 0.5 for a while.

dpsanders:
Closing this for now, but if you feel like using `@compat` with the new syntax, that would be helpful, @miguelraz, thanks.

ChrisRackauckas:
>but if you feel like using @compat with the new syntax

`@compat` with the new syntax won't work. That's because on v0.5 `struct ... end` doesn't parse, so it'll throw an error even with a macro. So you have to do `@compat type ... end` if you want v0.5 and v0.6 compatibility.

dpsanders:
Ah sorry, right. So we gain exactly nothing by doing so, apparently.

miguelraz:
We learned how to __not__ deprecate stuff in the future! Thanks for all the info, will keep an eye out.

ChrisRackauckas:
Found out it's Compat 0.17.0 that does this, not 0.18.0. Just for reference.

PR #262, opened 2017-03-19
Add bisection root-finder

### LINKED ISSUE #204: Add simple multi-dimensional root finder using bisection
cf. https://github.com/dpsanders/IntervalConstraintProgramming.jl/issues/64

dpsanders:
Fixes #204.

This adds a `bisection` root finder that works with single or multiple variables.

Currently, this involves modifying the `Root` type. 
For bisection in particular, this is, in some sense, not necessary, since it cannot prove uniqueness, so any roots are `:unknown`.

However, for multi-dimensional root finding in general, it is maybe necessary to do so.

### DIFF
--- a/src/root_finding/bisection.jl
+++ b/src/root_finding/bisection.jl
@@ -11,7 +11,7 @@ function bisection{T<:Union{Interval,IntervalBox}}(f, X::T; tolerance=1e-3, debu
     debug && @show X, image
 
     if !(zero(X) ⊆ image)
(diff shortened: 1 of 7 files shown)


dpsanders:
This is ready. Any objection to merging, @lbenet ?

dpsanders:
Moved to https://github.com/JuliaIntervals/IntervalRootFinding.jl/pull/7

PR #263, opened 2017-03-19
Add logic for geometric midpoint and checking dynamic range in bisect

dpsanders:


dpsanders:
I'm not sure this is actually necessary / a good idea. Don't merge yet.

dpsanders:
Closing since no longer relevant for this package.

PR #281, opened 2017-05-14
Fix badge

ChrisRackauckas:


lbenet:
Thanks for fixing this! I will merge this, since the problem with the tests is another story...

dpsanders:
It looks like all the badges are incorrect?

lbenet:
You are right. I've already fixed this locally.

Issue #293, opened 2023-01-31
Segmentation fault on Apple Silicon MacBook


iliailmer:
Could you please help me resolve this? The following code example leads to a segmentation fault on MacBook pro with M1 chip:

```julia
>]activate --temp
>]add ValidatedNumerics
>using ValidatedNumerics
>X = -4..4
```

Here are the error logs:

```
signal (11): Segmentation fault: 11
in expression starting at none:0
mpfr_vasnprintf_aux at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libmpfr.6.dylib (unknown line)
mpfr_snprintf at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libmpfr.6.dylib (unknown line)
round_string at /Users/iliailmer/.julia/packages/IntervalArithmetic/dhnJB/src/display.jl:161
round_string at /Users/iliailmer/.julia/packages/IntervalArithmetic/dhnJB/src/display.jl:172 [inlined]
basic_representation at /Users/iliailmer/.julia/packages/IntervalArithmetic/dhnJB/src/display.jl:190
representation at /Users/iliailmer/.julia/packages/IntervalArithmetic/dhnJB/src/display.jl:279 [inlined]
representation at /Users/iliailmer/.julia/packages/IntervalArithmetic/dhnJB/src/display.jl:279 [inlined]
show at /Users/iliailmer/.julia/packages/IntervalArithmetic/dhnJB/src/display.jl:339 [inlined]
show at ./multimedia.jl:47
unknown function (ip: 0x1315d00e3)
ijl_apply_generic at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libjulia-internal.1.8.dylib (unknown line)
#43 at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-macmini-aarch64-4.0/build/default-macmini-aarch64-4-0/julialang/julia-release-1-dot-8/usr/share/julia/stdlib/v1.8/REPL/src/REPL.jl:267
jfptr_YY.43_64233 at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/sys.dylib (unknown line)
ijl_apply_generic at /Applications/Julia-1.8.app/Contents/Resources/julia/lib/julia/libjulia-internal.1.8.dylib (unknown line)
[...]

PR #294, opened 2023-02-02
Resolve compat issues for m1 macs

iliailmer:
Resolves to #293

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -11,9 +11,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.5'
+          - '1'
           - '1.6'
-          - 'nightly'
         os:
           - ubuntu-latest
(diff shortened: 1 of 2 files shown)


lbenet:
Thanks for spotting the problems and open this PR.

I just run the tests, and there are problems apparently due to requirements from TaylorModels.jl with Julia version 1.5 (see https://github.com/JuliaIntervals/ValidatedNumerics.jl/actions/runs/4078312446/jobs/7046642429#step:5:29). Could you increase the Julia version required to Julia 1.6, as well as the tests, which should not consider Julia v1.5? We should also tag a new minor version?

iliailmer:
@lbenet I added the versions to CI.yml and Project.toml, should work now.

lbenet (approved):
LGTM! I'll go ahead and merge this, and also register the new version. Thanks a lot!

